PR #20, opened 2022-11-21
Netflix integration

skywojciechowskim:
Summary: Fixed errors in Netflix EME functions. Fixed logging in rialto server.
Type: Feature
Owner: Marcin Wojciechowski
Coding Standard Applied: Y
Test Plan: None
Jira: RIALTO-8

### DIFF
--- a/media/server/main/source/MediaKeySession.cpp
+++ b/media/server/main/source/MediaKeySession.cpp
@@ -199,15 +199,13 @@ MediaKeyErrorStatus MediaKeySession::closeKeySession()
         status = m_ocdmSession->cancelChallengeData();
         if (MediaKeyErrorStatus::OK != status)
         {
-            RIALTO_SERVER_LOG_ERROR("Failed to cancel the challenge data for the key session");
-            return status;
+            RIALTO_SERVER_LOG_WARN("Failed to cancel the challenge data for the key session");
         }
 
         status = m_ocdmSession->cleanDecryptContext();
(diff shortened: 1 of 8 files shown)


media/server/main/source/MediaKeySession.cpp
```diff
@@ -121,28 +122,18 @@ MediaKeyErrorStatus MediaKeySession::generateRequest(InitDataType initDataType,
...
     return MediaKeyErrorStatus::OK;
 }
 
 void MediaKeySession::getChallenge()
 {
+    RIALTO_SERVER_LOG_DEBUG("entry:");
     auto task = [&]()
     {
         uint32_t challengeSize = 0;
         MediaKeyErrorStatus status = m_ocdmSession->getChallengeData(m_kIsLDL, nullptr, &challengeSize);
```
skypetts, line 135: Why are we ignoring the return status?

skyrangelovv (approved):

media/server/main/source/MediaKeySession.cpp
skywojciechowskim: That's one of differences between OCDM for PR 2.5 and PR 4.0 :). As you can see, getChallengeData has to be called twice. First call is to find out what is the size of the challenge, second call is to get the challenge. First call for PR 4.0 OCDM always fails with ERROR_OUT_OF_MEMORY, it is also skipped in "real" netflix

skypetts (approved):

PR #36, opened 2022-12-12
Move decrypt after appsrc

skywilliamsonl:
Summary: Move decryption of samples after appsrc, to avoid allocating too many resources while storing decrypted buffers.
Type: Feature
Test Plan: Unittests, Youtube Certification Tests, Netflix Full Stack
Jira: RIALTO-10


media/server/gstplayer/source/GstSrc.cpp
```diff
@@ -375,8 +366,26 @@ void GstSrc::setupAndAddAppArc(GstElement *element, GstElement *appsrc, GstAppSr
     m_gstWrapper->gstBinAdd(GST_BIN(element), appsrc);
 
     GstElement *src_elem = appsrc;
+
+    // Configure and add decryptor
+    GstElement *decryptor = m_decryptorFactory->createDecryptorElement(nullptr, decryptionService);
+    if (decryptor)
+    {
+        GST_DEBUG("Injecting decryptor element %" GST_PTR_FORMAT, decryptor);
```
aczs, line 374: Can you change loggers in this function from _GST_XXX(...)_ to _GST_XXX_OBJECT(element, ...)_ ?

media/server/gstplayer/source/GstPlayer.cpp
```diff
@@ -390,9 +393,16 @@ GstBuffer *GstPlayer::createDecryptedBuffer(const IMediaPipeline::MediaSegment &
         }
         GstBuffer *subsamples = m_gstWrapper->gstBufferNewWrapped(subsamplesRaw, subsamplesRawSize);
 
-        m_decryptionService.decrypt(mediaSegment.getMediaKeySessionId(), gstBuffer, subsamples,
-                                    mediaSegment.getSubSamples().size(), initVector, keyId,
-                                    mediaSegment.getInitWithLast15());
+        GstStructure *info = m_gstWrapper->gstStructureNew("application/x-cenc", "encrypted", G_TYPE_BOOLEAN, TRUE,
```
aczs, line 396: Nit pick: I think that "encrypted" field is not needed since protection_meta will be present only when data is encrypted.
Name of the structure ("application/x-cenc") might be a little bit missleading, because we might want to use various enc scheme. Since only decrypter will use it, it can be any string, right?

skypetts: Agreed about the encrypted field, but I _think_ the name is probably actually a good change because of the upcoming OCDM changes to support cbcs. Currently only cenc is supported (as I understand it) and with the incoming changes we will introduce support for “application/x-cbcs” as well.

media/server/gstplayer/source/GstDecryptor.cpp
```diff
@@ -0,0 +1,298 @@
...
+    GST_DEBUG_CATEGORY_INIT(gst_rialto_decryptor_debug_category, "rialtodecryptor", 0, "Decryptor for Rialto");
+
+    GObjectClass *gobject_class = G_OBJECT_CLASS(klass);
+    gobject_class->finalize = GST_DEBUG_FUNCPTR(gst_rialto_decryptor_finalize);
+
+    GstElementClass *element_class = GST_ELEMENT_CLASS(klass);
+    gst_element_class_add_pad_template(element_class, gst_static_pad_template_get(&sinkTemplate));
+    gst_element_class_add_pad_template(element_class, gst_static_pad_template_get(&srcTemplate));
+    gst_element_class_set_static_metadata(element_class, "Rialto Decryptor", GST_ELEMENT_FACTORY_KLASS_DECRYPTOR,
+                                          "Decryptor for Rialto.", "Sky");
```
skypetts, line 82: Pedantic: The last field is supposed to be name & email isn't it? If you don't want to take all the blame you can add some others from the team in there too!

skypetts (commented):
Approving as all comments are minor

skypetts (dismissed):

media/server/gstplayer/source/GstPlayer.cpp
skywilliamsonl: I've changed the name to of the structure 'decryption_metadata' to make it more generic, we can add a new property to store the encryption scheme when we support multiple.

media/server/gstplayer/source/GstDecryptor.cpp
skywilliamsonl: You're right, added me and Adam.

media/server/gstplayer/source/GstSrc.cpp
skywilliamsonl: Done.

skypetts (approved):

aczs (approved):

PR #43, opened 2022-12-23
GStreamer Logs forwarding

skywojciechowskim:
Summary: GStreamer logs forwarded to rialto server logger.
Type: Feature
Owner: Marcin Wojciechowski
Coding Standard Applied: Y
Test Plan: None
Jira: RIALTO-47

### DIFF
--- a/logging/include/RialtoLogging.h
+++ b/logging/include/RialtoLogging.h
@@ -39,6 +39,8 @@ extern "C"
         RIALTO_DEBUG_LEVEL_MILESTONE = (1u << 3),
         RIALTO_DEBUG_LEVEL_INFO = (1u << 4),
         RIALTO_DEBUG_LEVEL_DEBUG = (1u << 5),
+        RIALTO_DEBUG_LEVEL_EXTERNAL =
+            (1u << 6), // Level controlled by an external variable, like e.g. GST_DEBUG. Keep it as last level.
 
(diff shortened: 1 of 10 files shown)


logging/include/RialtoLogging.h
```diff
@@ -39,6 +39,8 @@ extern "C"
         RIALTO_DEBUG_LEVEL_MILESTONE = (1u << 3),
         RIALTO_DEBUG_LEVEL_INFO = (1u << 4),
         RIALTO_DEBUG_LEVEL_DEBUG = (1u << 5),
+        RIALTO_DEBUG_LEVEL_EXTERNAL =
```
skypetts, line 42: Is there a reason why it needs to be last? If we want it in a fixed position why not add it add the start of the list?

skywojciechowskim: It's my suggestion only. It's last, because I'd like this level to be independent from RIALTO_DEBUG environment variable. Furthermore it's "unusual" level. I can remove comment in the code, if you don't like it.

logging/include/RialtoLogging.h
```diff
@@ -55,6 +57,7 @@ extern "C"
         RIALTO_COMPONENT_IPC,
         RIALTO_COMPONENT_SERVER_MANAGER,
         RIALTO_COMPONENT_COMMON,
+        RIALTO_COMPONENT_EXTERNAL, // External component, like e.g. GStreamer. Keep it as last component.
```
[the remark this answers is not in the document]
skywojciechowskim: It's also my suggestion only, it doesn't have to be last. I added this comment here, because EXTERNAL is not RIALTO component. I can also remove this comment if you don't like it.

skywilliamsonl (approved):

aczs (approved):

PR #53, opened 2023-01-04
Web audio server ipc and service implementation

skywilliamsonl:
Summary: Implements a new WebAudioPlayer module in ipc and service with Unittests. Also includes refactoring of PlaybackService so that the sharedMemory can be shared between MediaPipelines and WebAudioPlayers.
Type: Feature
Test Plan: Unittests
Jira: RIALTO-2


media/public/include/IWebAudioPlayer.h
```diff
@@ -155,6 +159,9 @@ class IWebAudioPlayer
      *
      * Sends a buffer of audio data for playback
      *
+     * A call to writeBuffer should always follow a getBufferAvailable. If no frames are written
```
skypetts, line 162: I thought we decided we didn't need this restriction?

media/server/ipc/source/WebAudioPlayerModuleService.cpp
```diff
@@ -0,0 +1,345 @@
...
+        if (request->config().has_pcm())
+        {
+            config.pcm.rate = request->config().pcm().rate();
+            config.pcm.channels = request->config().pcm().channels();
+            config.pcm.sampleSize = request->config().pcm().sample_size();
+            config.pcm.isBigEndian = request->config().pcm().is_big_endian();
+            config.pcm.isSigned = request->config().pcm().is_signed();
+            config.pcm.isFloat = request->config().pcm().is_float();
+        }
+    }
```
skypetts, line 135: What happens if there is no config or PCM parameters? For now, I suppose we should return an error in such a case.

skywilliamsonl: Yes agreed, however I will add the logic to WebAudioPlayer, not the service layer.

skypetts (approved):

skywojciechowskim (approved):

PR #63, opened 2023-01-30
SourceId dynamic assignment

skywojciechowskim:
Summary: SourceId dynamic assignment
Type: Feature
Owner: Marcin Wojciechowski
Coding Standard Applied: Y
Test Plan: None
Jira: RIALTO-28

### DIFF
--- a/media/client/main/source/MediaPipeline.cpp
+++ b/media/client/main/source/MediaPipeline.cpp
@@ -140,8 +140,7 @@ MediaPipeline::MediaPipeline(std::weak_ptr<IMediaPipelineClient> client, const V
                              const std::shared_ptr<IMediaPipelineIpcFactory> &mediaPipelineIpcFactory,
                              const std::shared_ptr<common::IMediaFrameWriterFactory> &mediaFrameWriterFactory,
                              const std::shared_ptr<ISharedMemoryManagerFactory> &sharedMemoryManagerFactory)
-    : m_mediaPipelineClient(client), m_mediaFrameWriterFactory(mediaFrameWriterFactory),
-      m_currentState(State::IDLE), m_audioSourceSwitchOngoing{false}
+    : m_mediaPipelineClient(client), m_mediaFrameWriterFactory(mediaFrameWriterFactory), m_currentState(State::IDLE)
 {
     RIALTO_CLIENT_LOG_DEBUG("entry:");
 
@@ -195,11 +194,7 @@ bool MediaPipeline::attachSource(const std::unique_ptr<IMediaPipeline::MediaSour
     if (status)
     {
         source->setId(sourceId);
-    }
-    if (m_audioSourceSwitchOngoing && MediaSourceType::AUDIO == source->getType())
-    {
-        RIALTO_CLIENT_LOG_DEBUG("Clearing Audio Source Switch Ongoing flag");
-        m_audioSourceSwitchOngoing = false;
+        m_attachedSources.add(sourceId, source->getType());
     }
     return status;
 }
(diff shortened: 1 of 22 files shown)


media/client/main/source/MediaPipeline.cpp
```diff
@@ -566,9 +555,10 @@ void MediaPipeline::notifyNeedMediaData(int32_t sourceId, size_t frameCount, uin
 {
     RIALTO_CLIENT_LOG_DEBUG("entry:");
 
-    if (m_audioSourceSwitchOngoing && MediaSourceType::AUDIO == static_cast<MediaSourceType>(sourceId))
+    if (MediaSourceType::UNKNOWN == m_attachedSources.get(sourceId))
```
skypetts, line 558: This is kinda nice but at the same time I think there's always the edge case that the app may get a spurious needData for a removed source depending on timing & thread model so it will need some similar logic to this anyway. So there would be an argument for keeping Rialto client simple and removing this check. I don't feel strongly either way personally - what do others think?

media/server/main/source/MediaPipelineServerInternal.cpp
```diff
@@ -243,7 +249,20 @@ bool MediaPipelineServerInternal::attachSourceInternal(const std::unique_ptr<Med
     }
 
     m_gstPlayer->attachSource(source);
-    source->setId(static_cast<int32_t>(source->getType()));
+
+    const auto kSourceIter = m_attachedSources.find(source->getType());
+    if (m_attachedSources.cend() == kSourceIter)
+    {
+        source->setId(generateSourceId());
```
skypetts, line 256: Maybe I'm missing something, but why don't we use the sourceId passed in by the client in the MediaSoruce object?

media/server/gstplayer/source/tasks/generic/AttachSource.cpp
```diff
@@ -311,8 +311,8 @@ void AttachSource::switchAudioSource(GstCaps *caps, const std::string &strCaps)
         m_gstWrapper->gstCapsUnref(oldCaps);
     }
     RIALTO_SERVER_LOG_MIL("Switching audio source.");
-    RIALTO_SERVER_LOG_DEBUG("Old caps: %s", oldCapsStr.c_str());
-    RIALTO_SERVER_LOG_DEBUG("New caps: %s", strCaps.c_str());
+    RIALTO_SERVER_LOG_MIL("Old caps: %s", oldCapsStr.c_str());
```
skypetts, line 314: I think these should either stay as _DEBUG or combine all 3 log lines into a single _MIL so as not to produce excessive MIL log lines

media/server/main/source/MediaPipelineServerInternal.cpp
skywojciechowskim: Source ID is generated by Server, Client doesn't set this value in AttachSourceRequest, sourceId is a part of AttachSourceResponse:
message AttachSourceResponse {
    required int32 source_id = 1;
}

media/client/main/source/MediaPipeline.cpp
skywojciechowskim: Rialto sink checks, if streamId is still attached before pulling sample, so I can remove this check. In such case developers of apps using rialto c++ interface will have to remember to skip attaching segments for removed source ids.

skypetts: As discussed we will keep this

skypetts (approved):

aczs (approved):

PR #69, opened 2023-02-15
Set codec data

skywojciechowskim:
Summary: SetCodecData API added
Type: Fix
Owner: Marcin Wojciechowski
Coding Standard Applied: Y
Test Plan: None
Jira: RIALTO-66

### DIFF
--- a/media/public/include/IMediaPipeline.h
+++ b/media/public/include/IMediaPipeline.h
@@ -136,7 +136,7 @@ public:
         /**
          * @brief Gets the codec data
          */
-        const std::vector<uint8_t> &getCodecData() const { return m_codecData; }
+        const CodecData &getCodecData() const { return m_codecData; }
 
         /**
          * @brief Gets the stream format
@@ -158,7 +158,7 @@ public:
                              const std::string &mimeType = std::string(),
                              SegmentAlignment alignment = SegmentAlignment::UNDEFINED,
                              StreamFormat streamFormat = StreamFormat::UNDEFINED,
-                             const std::vector<uint8_t> &codecData = std::vector<uint8_t>())
+                             const CodecData &codecData = CodecData())
             : m_id(id), m_configType(configType), m_mimeType(mimeType), m_alignment(alignment),
               m_streamFormat(streamFormat), m_codecData(codecData)
         {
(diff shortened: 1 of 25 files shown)


media/public/include/MediaCommon.h
```diff
@@ -332,6 +332,56 @@ enum class WebAudioPlayerState
     FAILURE        /**< The player failed to set playback state. */
 };
 
+/**
+ * @brief Additional data for decoder.
+ */
+class CodecData
```
skypetts, line 338: This file was initially design to hold some simple data types likely to be used across the system, are we happy adding a class in the public interface here?

I don't have any issue with something like this internally in Rialto server but wonder if we could keep things cleaner for the client by leaving the public interface using vector<uint8_t> and perhaps a private flag in the media segment class to indicate if the value is set.

@aczs & @skywilliamsonl what do you think?

media/public/include/IMediaPipeline.h
```diff
@@ -578,6 +587,15 @@ class IMediaPipeline
          */
         void setSegmentAlignment(const SegmentAlignment &alignment) { m_alignment = alignment; }
 
+        /**
+         * @brief Sets new codec_data for the source.
```
skypetts, line 591: source->segment

media/public/include/MediaCommon.h
skywilliamsonl: Yeah, in my opinion it clutters this header.

skywojciechowskim: vector<uint8_t> + flag in IMediaPipeline interface is not a good option in my opinion, because we would have to add both params to MediaSource constructor. I will change this class to a simple struct with vector and bool to keep both params in one place, user will have to remember to set both params when object is created.

skywojciechowskim: Another option is to move current class to IMediaPipeline class (like MediaSource). What do You prefer?

media/public/include/IMediaPipeline.h
skywojciechowskim: I will change it.

media/public/include/MediaCommon.h
skywojciechowskim: As agreed on Slack, I will change it to std::shared_ptr<std::vector<uint8_t>>.

skypetts (approved):

skywilliamsonl (approved):

PR #78, opened 2023-02-24
Improve rialto for multithreading

skywilliamsonl:
Summary: Add a new mutex in RialtoControl to protect from multiple threads changing state at the same time and add a mutex to singleton objects to protect from two threads creating the first object at the same time.
Type: Cleanup
Test Plan: Unittests, Youtube Cert Tests
Jira: RIALTO-15


media/client/ipc/source/MediaKeysCapabilitiesIpc.cpp
```diff
@@ -42,6 +43,8 @@ std::shared_ptr<IMediaKeysCapabilitiesIpcFactory> IMediaKeysCapabilitiesIpcFacto
 
 std::shared_ptr<IMediaKeysCapabilities> MediaKeysCapabilitiesIpcFactory::getMediaKeysCapabilitiesIpc() const
 {
+    std::lock_guard<std::mutex> lock{m_creationMutex};
```
skywojciechowskim, line 46: You won't need those creation mutexes if You use Meyer's singleton. It's generally thread-safe:
https://laristra.github.io/flecsi/src/developer-guide/patterns/meyers_singleton.html

skypetts: I like this but I don't think it's a priority right now. Let's deliver this as is if there are no other comments and I've raised https://jira.rdkcentral.com/jira/browse/RIALTO-115 to make this improvement.

skypetts (approved):

skywojciechowskim (approved):

PR #87, opened 2023-03-07
Rialto Logger

skyshaha1:
Summary: RialtoLogger Improvement
Type: Fix
Test Plan: Tested via build_ut and check compilation via bitbake


logging/source/RialtoLogging.cpp
```diff
@@ -42,91 +42,129 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    case RIALTO_COMPONENT_IPC:
+        return "IPC";
+        break;
+    case RIALTO_COMPONENT_SERVER_MANAGER:
+        return "SMG";
+        break;
+    case RIALTO_COMPONENT_COMMON:
+        return "COM";
+        break;
+    case RIALTO_COMPONENT_LAST:
```
aczs, line 75: No need to convert 'component' RIALTO_COMPONENT_LAST to string. It's used only to know how many enum values we have.

logging/source/RialtoLogging.cpp
```diff
@@ -42,91 +42,129 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+        return "LST";
+        break;
+    case RIALTO_COMPONENT_EXTERNAL:
+    default:
+        return "EXT";
```
aczs, line 80: Maybe default value should be something like "UNK" (from unknown)

logging/source/RialtoLogging.cpp
```diff
@@ -42,91 +42,129 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
     case RIALTO_DEBUG_LEVEL_DEBUG:
-        iov[1].iov_base = const_cast<void *>(reinterpret_cast<const void *>("DBG: "));
-        iov[1].iov_len = 5;
+        return "DBG";
         break;
     case RIALTO_DEBUG_LEVEL_EXTERNAL:
     default:
-        iov[1].iov_base = const_cast<void *>(reinterpret_cast<const void *>(": "));
-        iov[1].iov_len = 2;
+        return "EXT";
```
aczs, line 109: think that value form EXTERNAL should be EXT, but for default should be ": "

logging/source/RialtoLogging.cpp
```diff
@@ -42,91 +42,129 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    struct iovec iov[6];
+    char tbuf[32]; 
+
+    iov[0].iov_base = tbuf;
+    iov[0].iov_len = snprintf(tbuf, sizeof(tbuf), "%.010lu.%.06lu ", ts.tv_sec, ts.tv_nsec / 1000);
+    iov[0].iov_len = std::min(iov[0].iov_len, sizeof(tbuf));
+
+    char lbuf[180];
+    iov[1].iov_base = reinterpret_cast<void *>(lbuf);
+    iov[1].iov_len = snprintf(lbuf, sizeof(lbuf), "%s: ", levelToString(level).c_str());
```
aczs, line 131: nit pick: snprintf returns "The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character."
Meaning that we could have buffer overflow if _level_ string would be too long (it will not be with our current level names). I would use _std::min_ here as it's done with all snprintfs in _consoleLogHandle_r function just in case.

logging/source/RialtoLogging.cpp
```diff
@@ -42,91 +42,129 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+
+    char cbuf[180];
+    iov[2].iov_base = reinterpret_cast<void *>(cbuf);
+    iov[2].iov_len = snprintf(cbuf, sizeof(cbuf), "%s: ", componentToString(component).c_str());
```
aczs, line 135: same here

logging/source/RialtoLogging.cpp
```diff
@@ -42,91 +42,129 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    timespec ts = {0, 0};
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    struct iovec iov[6];
+    char tbuf[32]; 
+
+    iov[0].iov_base = tbuf;
+    iov[0].iov_len = snprintf(tbuf, sizeof(tbuf), "%.010lu.%.06lu ", ts.tv_sec, ts.tv_nsec / 1000);
+    iov[0].iov_len = std::min(iov[0].iov_len, sizeof(tbuf));
+
+    char lbuf[180];
```
aczs, line 129: Maybe make this buf smaller (8 chars?)

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,148 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    case RIALTO_COMPONENT_IPC:
+        return "IPC";
+        break;
+    case RIALTO_COMPONENT_SERVER_MANAGER:
+        return "SMG";
+        break;
+    case RIALTO_COMPONENT_COMMON:
+        return "COM";
+        break;
+    case RIALTO_COMPONENT_EXTERNAL:
```
aczs, line 75: For RIALTO_COMPONENT_EXTERNAL it should be "EXT" and default stay "UNK"

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,148 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    struct iovec iov[6];
+    char tbuf[32];
+
+    iov[0].iov_base = tbuf;
+    iov[0].iov_len = snprintf(tbuf, sizeof(tbuf), "%.010lu.%.06lu ", ts.tv_sec, ts.tv_nsec / 1000);
+    iov[0].iov_len = std::min(iov[0].iov_len, sizeof(tbuf));
+
+    char lbuf[8];
+    iov[1].iov_base = reinterpret_cast<void *>(lbuf);
+    iov[1].iov_len = std::min(snprintf(lbuf, sizeof(lbuf), "%s: ", levelToString(level).c_str()), 8);
```
aczs, line 130: replace 8 with sizeof(lbuf) for maintainability

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,148 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+
+    char cbuf[8];
+    iov[2].iov_base = reinterpret_cast<void *>(cbuf);
+    iov[2].iov_len = std::min(snprintf(cbuf, sizeof(cbuf), "%s: ", componentToString(component).c_str()), 8);
```
aczs, line 134: replace 8 with sizeof(cbuf) for maintainability

skyshaha1: So I tried changing it initially with sizeof(cbuf), then when I ran ./build_ut.py -c -s client, I get an error (
`note:   deduced conflicting types for parameter ‘const _Tp’ (‘int’ and ‘long unsigned int’)
  134 |     iov[2].iov_len = std::min(snprintf(cbuf, sizeof(cbuf), "%s: ", componentToString(component).c_str()), sizeof(cbuf));`)

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,150 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    char tbuf[32];
+
+    iov[0].iov_base = tbuf;
+    iov[0].iov_len = snprintf(tbuf, sizeof(tbuf), "%.010lu.%.06lu ", ts.tv_sec, ts.tv_nsec / 1000);
+    iov[0].iov_len = std::min(iov[0].iov_len, sizeof(tbuf));
+
+    char lbuf[8];
+    iov[1].iov_base = reinterpret_cast<void *>(lbuf);
+    iov[1].iov_len =
+        std::min(snprintf(lbuf, sizeof(lbuf), "%s: ", levelToString(level).c_str()), static_cast<int>(sizeof(lbuf)));
```
aczs, line 131: I think it's better to cast snprintf result to size_t, since iov_len is also size_t.

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,150 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+
+    char cbuf[8];
+    iov[2].iov_base = reinterpret_cast<void *>(cbuf);
+    iov[2].iov_len = std::min(snprintf(cbuf, sizeof(cbuf), "%s: ", componentToString(component).c_str()),
```
aczs, line 135: I think it's better to cast snprintf result to size_t, since iov_len is also size_t.

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,154 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+        return "IPC";
+        break;
+    case RIALTO_COMPONENT_SERVER_MANAGER:
+        return "SMG";
+        break;
+    case RIALTO_COMPONENT_COMMON:
+        return "COM";
+        break;
+    case RIALTO_COMPONENT_EXTERNAL:
+        return "EXT:";
```
aczs, line 76: Remove ':'

aczs (dismissed):

logging/source/RialtoLogging.cpp
```diff
@@ -42,108 +42,154 @@ static const firebolt::rialto::logging::EnvVariableParser g_envVariableParser;
...
+    char tbuf[32];
+
+    iov[0].iov_base = tbuf;
+    iov[0].iov_len = snprintf(tbuf, sizeof(tbuf), "%.010lu.%.06lu ", ts.tv_sec, ts.tv_nsec / 1000);
+    iov[0].iov_len = std::min(iov[0].iov_len, sizeof(tbuf));
+
+    char lbuf[8];
+    iov[1].iov_base = reinterpret_cast<void *>(lbuf);
+    iov[1].iov_len =
+        std::min(static_cast<size_t>(snprintf(lbuf, sizeof(lbuf), "%s: ", componentToString(component).c_str())),
```
skypetts, line 133: Should this be using levelToString() rather than componentToString()? It looks like a duplicate of what's sent in iov[2] at the moment.

skypetts (approved):

aczs (approved):

PR #105, opened 2023-03-30
Fix NPLB tests

skywilliamsonl:
Summary: Avoid sending unneeded need datas when an EOS has been received & detatch IPC when there is a failure to create objects over IPC.
Type: Bugfix
Test Plan: Unittests, NPLB, Youtube Cert Tests
Jira: RIALTO-118 & RIALTO-123

### DIFF
--- a/media/server/main/include/MediaPipelineServerInternal.h
+++ b/media/server/main/include/MediaPipelineServerInternal.h
@@ -224,9 +224,9 @@ protected:
     bool m_wasAllSourcesAttachedCalled;
 
     /**
-     * @brief Flag used to check if Eos has been set on this playback
+     * @brief Map of flag used to check if Eos has been set on the media type for this playback
      */
-    bool m_isEos;
+    std::map<MediaSourceType, bool> m_isMediaTypeEosMap;
 
     /**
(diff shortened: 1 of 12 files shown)


media/server/main/include/MediaPipelineServerInternal.h
```diff
@@ -223,6 +223,11 @@ class MediaPipelineServerInternal : public IMediaPipelineServerInternal, public
      */
     bool m_wasAllSourcesAttachedCalled;
 
+    /**
+     * @brief Flag used to check if Eos has been set on this playback
+     */
+    bool m_isEos;
```
skywojciechowskim, line 229: Just a question - shouldn't we have m_isEos for each media type? What if we reach EOS for video, but we still need some data for audio?

skywilliamsonl: Yeah you are right, will change that.

skywojciechowskim (dismissed):

aczs (approved):

skywojciechowskim (approved):

aczs (approved):

PR #113, opened 2023-04-25
Env variables substituted with params in Rialto Server Manager so

skywojciechowskim:
Summary: Env variables substituted with params in Rialto Server Manager so
Type: Feature
Test Plan: UT
Jira: RIALTO-4

### DIFF
--- a/common/public/include/SessionServerCommon.h
+++ b/common/public/include/SessionServerCommon.h
@@ -73,9 +73,12 @@ struct AppConfig
  */
 struct ServerManagerConfig
 {
-    std::list<std::string> sessionServerEnvVars{}; /* List of environment variables, that need to be passed to
-                                                      RialtoSessionServer */
-    int numOfPreloadedServers{0};                  /* Number of preloaded servers */
+    std::list<std::string> sessionServerEnvVars{};          /* List of environment variables, that need to be passed to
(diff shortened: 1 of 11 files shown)


common/public/include/SessionServerCommon.h
```diff
@@ -73,9 +74,12 @@ struct AppConfig
  */
 struct ServerManagerConfig
 {
-    std::list<std::string> sessionServerEnvVars{}; /* List of environment variables, that need to be passed to
-                                                      RialtoSessionServer */
-    int numOfPreloadedServers{0};                  /* Number of preloaded servers */
+    std::list<std::string> sessionServerEnvVars{};          /* List of environment variables, that need to be passed to
+                                                               RialtoSessionServer */
+    int numOfPreloadedServers{0};                           /* Number of preloaded servers */
```
aczs, line 79: nit pick: I think that uint would be better type for numOfPreloadedServers

aczs (approved):

common/public/include/SessionServerCommon.h
skywojciechowskim: done

skywilliamsonl (approved):

aczs (approved):

PR #135, opened 2023-06-08
Underflow remake

aczs:


skywojciechowskim (approved):

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -678,16 +678,6 @@ void GstGenericPlayer::cancelUnderflow(bool &underflowFlag)
         return;
     }
     underflowFlag = false;
-    if (!m_context.audioUnderflowOccured && !m_context.videoUnderflowOccured)
```
skywilliamsonl, line 681: Do we need these flags and cancel underflow anymore? Seems all we do is set now, so I think we can tidy this up. Can get this delivered first as it is though.

skywilliamsonl (approved):

PR #189, opened 2023-09-27
Change all required parameters to optional

skyshaha1:
Summary: Change all required parameters to optional
Type: Fix
Test Plan: Unit Tests & Full Stack
Jira: RIALTO-303

### DIFF
--- a/proto/mediapipelinemodule.proto
+++ b/proto/mediapipelinemodule.proto
@@ -44,11 +44,11 @@ option cc_generic_services = true;
  * @returns a unique numeric session id value that should be used for all subsequent operations on the session.
  */
 message CreateSessionRequest {
-    required uint32 max_width = 1;
-    required uint32 max_height = 2;
+    optional uint32 max_width = 1 [default = 0];
+    optional uint32 max_height = 2 [default = 0];
 }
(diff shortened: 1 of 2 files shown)


proto/mediapipelinemodule.proto
```diff
@@ -82,10 +82,10 @@ message LoadRequest {
         MSE = 1;            ///< Media is MSE and will request data.
     }
 
-    required int32      session_id = 1;
-    required MediaType  type = 2;
-    required string     mime_type = 3;
-    required string     url = 4;
+    optional int32      session_id = 1 [default = -1];
+    optional MediaType  type = 2 [default = UNKNOWN];
+    optional string     mime_type = 3 [default = " "];
```
skywojciechowskim, line 87: why space? Shouldn't it be just empty string?

proto/mediapipelinemodule.proto
```diff
@@ -82,10 +82,10 @@ message LoadRequest {
...
+    optional string     url = 4 [default = " "];
```
skywojciechowskim, line 88: same as above

skywojciechowskim (approved):

PR #201, opened 2023-10-25
Rialto 307 change owner and group of socket

skyproudmanp:
Summary: Add settings which allow specifying owner and group of the rialto socket
Type: Feature
Test Plan: Ran unit tests. Loaded the image onto Sky0 box and changed the appsservice code to specify a different owner and group for the socket and the socket was created as expected. Also, a test was conducted with the standard appservice code (not specifying an owner and group... default empty strings) and the socket was created as expected (with the user and primary group running the process)
Dependencies and Impacts: Extra parameters for socket owner and group are added alongside the pre-existing socket permissions
Jira: RIALTO-307

To change the appsservice so that it specifies an owner/ group...

RDK/AppManager/RialtoServerManager/source/RialtoServerManager.cpp
@@ -59,6 +59,10 @@ RialtoServerManager::RialtoServerManager(const std::list<std::string> &environme
     firebolt::rialto::common::ServerManagerConfig config;
     config.sessionServerEnvVars = addMandatoryEnvVars(environmentVars);
     config.numOfPreloadedServers = numOfPreloadedServers;
+
+    config.sessionManagementSocketPermissions.owner = "app-500";
+    config.sessionManagementSocketPermissions.group = "apps";
+
     mServerManagerService = rialto::servermanager::service::create(mStateObserver, config);
     if (!mServerManagerService->registerLogHandler(mLogHandler))
     {


proto/servermanagermodule.proto
```diff
@@ -88,6 +90,8 @@ message SetConfigurationRequest {
     optional LogLevels logLevels = 4;
     optional uint32 socketPermissions = 5;
     optional string clientDisplayName = 6;
+    required string socketOwner = 7;
```
skywojciechowskim, line 93: optional

proto/servermanagermodule.proto
```diff
@@ -88,6 +90,8 @@ message SetConfigurationRequest {
...
+    required string socketGroup = 8;
```
skywojciechowskim, line 94: optional

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
     {
         RIALTO_SERVER_LOG_SYS_WARN(errno, "Failed to change the permissions on the IPC socket");
     }
+
+    const uid_t noOwnerChange = -1; // -1 means chown() won't change the owner
```
skywojciechowskim, line 92: constexpr, according to our naming policy, const values should start with k, in this case kNoOwnerPolicy

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
...
+        {
+            RIALTO_SERVER_LOG_SYS_WARN(errno, "Failed to determine ownerId for '%s'", socketOwner.c_str());
+        }
+        else
+        {
+            ownerId = passwordResult->pw_uid;
+        }
+    }
+
+    const gid_t noGroupChange = -1; // -1 means chown() won't change the group
```
skywojciechowskim, line 112: constexpr, name

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
     {
         RIALTO_SERVER_LOG_SYS_WARN(errno, "Failed to change the permissions on the IPC socket");
     }
+
```
skywojciechowskim, line 91: This function became very long, I would prefer to have 2 functions here, uid_t getOwnerId(...) and gid_t getGroupId(...). chown at the end can stay where it is. You can place common constants (noOwnerChange, noGroupChange) in empty namespace at the top of this file (please look at other files as an example)

tests/media/server/ipc/serverManagerModuleService/ServerManagerModuleServiceTestsFixture.cpp
```diff
@@ -33,6 +33,8 @@ constexpr int MAX_SESSIONS{5};
 constexpr int MAX_WEB_AUDIO_PLAYERS{3};
 constexpr int socket{2};
 constexpr int pingId{29};
+const std::string SOCKET_OWNER;
```
skywojciechowskim, line 36: please define some value, like "root", not empty strings

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.cpp
```diff
@@ -42,6 +42,8 @@ namespace
 {
 const std::string socketName{"/tmp/sessionmanagementservertest-0"};
 constexpr unsigned int socketPermissions{0777};
+const std::string socketOwner;
```
skywojciechowskim, line 45: define value

tests/media/server/service/sessionServerManager/SessionServerManagerTestsFixture.cpp
```diff
@@ -45,6 +45,8 @@ constexpr firebolt::rialto::common::MaxResourceCapabilitites maxResource{maxPlay
 constexpr RIALTO_DEBUG_LEVEL logLvl{RIALTO_DEBUG_LEVEL_DEFAULT};
 const std::string sessionManagementSocket{"/tmp/rialtosessionservermanagertests-0"};
 constexpr unsigned int sessionManagementSocketPermissions{0777};
+const std::string sessionManagementSocketOwner;
```
skywojciechowskim, line 48: define value

tests/serverManager/unittests/common/SessionServerAppTestsFixture.cpp
```diff
@@ -30,6 +30,8 @@ const std::string kSessionServerPath{"/usr/bin/RialtoServer"};
 constexpr std::chrono::milliseconds kSessionServerStartupTimeout{100};
 constexpr std::chrono::milliseconds kKillTimeout{1000};
 constexpr unsigned int kSocketPermissions{0777};
+const std::string kSocketOwner;
```
skywojciechowskim, line 33: define value

tests/serverManager/unittests/ipc/IpcTestsFixture.cpp
```diff
@@ -135,8 +135,10 @@ bool IpcTests::triggerPerformSetConfiguration()
     const std::string clientSocketName{"westeros-rialto"};
     constexpr firebolt::rialto::common::MaxResourceCapabilitites maxResource{2, 1};
     constexpr unsigned int kSocketPermissions{0777};
+    const std::string kSocketOwner;
```
skywojciechowskim, line 138: define value

common/public/include/SessionServerCommon.h
```diff
@@ -78,13 +78,15 @@ constexpr unsigned int kWrite{2};
...
 /**
- * @brief Socket permissions
+ * @brief Socket permissions and ownership
  */
 struct SocketPermissions
 {
     unsigned int ownerPermissions{kRead | kWrite};
     unsigned int groupPermissions{kRead | kWrite};
     unsigned int otherPermissions{kRead | kWrite};
+    std::string owner;
```
skywojciechowskim, line 88: Define default values. In my opinion they should be root/root. Unfortunately You missed one functionality, you have to add it - recently Adam has introduced ConfigReader (in server manager). Your new values have to be defined in json config file. Please look at Adam's change to know the details.

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
...
+    const uid_t noOwnerChange = -1; // -1 means chown() won't change the owner
+    uid_t ownerId = noOwnerChange;
+    if (!socketOwner.empty())
+    {
+        errno = 0;
+        struct passwd passwordStruct;
+        struct passwd *passwordResult;
+        char stringBuffer[1024];
```
skywilliamsonl, line 99: magic number, add constexpr variable for buffer size and use variable for the bufsize in getpwnam_r

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
...
+        getpwnam_r(socketOwner.c_str(), &passwordStruct, stringBuffer, sizeof(stringBuffer) / sizeof(*stringBuffer),
```
skywilliamsonl, line 100: getpwnam_r returns int, check the return value.

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
...
         RIALTO_SERVER_LOG_SYS_WARN(errno, "Failed to change the permissions on the IPC socket");
     }
+
+    const uid_t noOwnerChange = -1; // -1 means chown() won't change the owner
+    uid_t ownerId = noOwnerChange;
+    if (!socketOwner.empty())
+    {
+        errno = 0;
+        struct passwd passwordStruct;
+        struct passwd *passwordResult;
```
skywilliamsonl, line 98: Initialise to nullptr

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
...
+        char stringBuffer[1024];
+        getpwnam_r(socketOwner.c_str(), &passwordStruct, stringBuffer, sizeof(stringBuffer) / sizeof(*stringBuffer),
+                   &passwordResult);
+        if (passwordResult == NULL)
```
skywilliamsonl, line 102: Use nullptr

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
     {
         RIALTO_SERVER_LOG_SYS_WARN(errno, "Failed to change the permissions on the IPC socket");
     }
+
+    const uid_t noOwnerChange = -1; // -1 means chown() won't change the owner
+    uid_t ownerId = noOwnerChange;
+    if (!socketOwner.empty())
+    {
+        errno = 0;
+        struct passwd passwordStruct;
```
skywilliamsonl, line 97: Default initialise {}

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -82,6 +88,56 @@ bool SessionManagementServer::initialize(const std::string &socketName, unsigned
...
+        {
+            ownerId = passwordResult->pw_uid;
+        }
+    }
+
+    const gid_t noGroupChange = -1; // -1 means chown() won't change the group
+    gid_t groupId = noGroupChange;
+    if (!socketGroup.empty())
+    {
+        errno = 0;
```
skywilliamsonl, line 116: Comments above are also relevant here

skywojciechowskim:
> @skyproudmanp generally its going to be difficult to unittests these linux functions, maybe we should create a new wrapper in media/server/wrappers for these and then we could mock them properly. Can do that as a separate PR if you'd like, rather than lump it in here.

You don't have to create new wrapper - we already have one in common directory -> rialto/common/interface/ILinuxWrapper.h

common/public/include/SessionServerCommon.h
skyproudmanp: I was thinking that it would be better to have the default values as empty strings, and then the "SessionManagementServer::initialize" method does not call "chown". This will leave the created socket being owned by the user executing the code (and the group would be their primary group). Perhaps I could add a comment to make this intention clear... or do you still think that I should have root/root as default values?

skywojciechowskim: ok, that's fine, please add a comment.

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -25,10 +25,78 @@
...
+#include <unistd.h>
 
+namespace
+{
+const uid_t kNoOwnerChange = -1; // -1 means chown() won't change the owner
+const gid_t kNoGroupChange = -1; // -1 means chown() won't change the group
+} // namespace
 namespace firebolt::rialto::server::ipc
 {
+static size_t getBufferSizeForPasswordStructureCalls()
```
skywojciechowskim, line 41: placing functions in empty namespaces instead of static functions is more c++ style :)

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTests.cpp
```diff
@@ -49,6 +49,26 @@ TEST_F(SessionManagementServerTests, shouldEstablishConnection)
     sendConnectClient();
 }
 
+TEST_F(SessionManagementServerTests, testSocketOwnershipValidOwnerAndGroup)
+{
+    testSocketOwnership(true, true);
```
skywilliamsonl, line 54: Usually we split the test cases up into expect calls and invocation.

media/server/ipc/include/SessionManagementServer.h
```diff
@@ -58,7 +58,9 @@ class SessionManagementServer : public ISessionManagementServer
     SessionManagementServer &operator=(const SessionManagementServer &) = delete;
     SessionManagementServer &operator=(SessionManagementServer &&) = delete;
 
-    bool initialize(const std::string &socketName, unsigned int socketPermissions) override;
+    bool initialize(std::unique_ptr<firebolt::rialto::common::ILinuxWrapper> &linuxWrapper,
```
skywojciechowskim, line 61: In most of cases in rialto wrappers are passed in class constructor. Please move it there to make code more unified.

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.cpp
```diff
@@ -184,12 +186,57 @@ void SessionManagementServerTests::serverWillSetLogLevels()
...
 
 void SessionManagementServerTests::sendServerInitializeAndExpectFailure()
 {
-    EXPECT_FALSE(m_sut->initialize(socketName, socketPermissions, socketOwner, socketGroup));
+    std::unique_ptr<firebolt::rialto::common::ILinuxWrapper> linuxWrapper =
+        firebolt::rialto::common::ILinuxWrapperFactory::createFactory()->createLinuxWrapper();
+    EXPECT_FALSE(m_sut->initialize(linuxWrapper, socketName, socketPermissions, socketOwner, socketGroup));
+}
+
+void SessionManagementServerTests::testSocketOwnership(bool testOwnerIsValid, bool testGroupIsValid)
```
skywilliamsonl, line 201: Split this method up, methods should have minimal logic so that the test can be followed clearly.

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -25,10 +25,78 @@
...
+{
+    // Can return -1 on error
+    return sysconf(_SC_GETPW_R_SIZE_MAX);
+}
+
+static uid_t getSocketOwnerId(std::unique_ptr<firebolt::rialto::common::ILinuxWrapper> &linuxWrapper,
```
skywojciechowskim, line 47: If you pass LinuxWrapper in class constructor, it will be private member attribute. In such case this function and next one (getSocketGroupId) can be private const class member method.

media/server/ipc/source/SessionManagementServer.cpp
skywojciechowskim: it looks perfect now! :)

media/server/ipc/source/SessionManagementServer.cpp
skyproudmanp: Thanks, I didn't know this

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -31,8 +31,72 @@
...
+                              const std::string &kSocketOwner)
+{
+    uid_t ownerId = kNoOwnerChange;
+    const size_t kBufferSize = getBufferSizeForPasswordStructureCalls();
+    if (!kSocketOwner.empty() && kBufferSize > 0)
+    {
+        errno = 0;
+        struct passwd passwordStruct
```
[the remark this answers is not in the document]
skyproudmanp: Yes, clang did this :-/

media/server/ipc/source/SessionManagementServer.cpp
```diff
@@ -63,7 +127,8 @@ SessionManagementServer::~SessionManagementServer()
     }
 }
 
-bool SessionManagementServer::initialize(const std::string &socketName, unsigned int socketPermissions,
+bool SessionManagementServer::initialize(std::unique_ptr<firebolt::rialto::common::ILinuxWrapper> &linuxWrapper,
```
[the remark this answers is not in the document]
skyproudmanp: Cool, I'll do this. But, it seems a bit of a shame to increase the scope to being a class wide attribute when it's only needed by one method.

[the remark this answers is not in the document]
skywilliamsonl: I agree, but keeping it in the factory/constructor removes it from the interface so keeps it local to this class.

skywojciechowskim (approved):

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.cpp
```diff
@@ -182,12 +194,107 @@ void SessionManagementServerTests::serverWillSetLogLevels()
 
 void SessionManagementServerTests::sendServerInitialize()
 {
-    EXPECT_TRUE(m_sut->initialize(socketName, socketPermissions));
+    EXPECT_CALL(*m_linuxWrapperMock, chmod(testing::StrEq(socketName), socketPermissions)).WillOnce(Return(0));
```
skywilliamsonl, line 197: This EXPECT_CALL should be in serverWillInitialize.

We usually separate expect calls and invocation.

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.cpp
```diff
@@ -41,11 +43,19 @@ using testing::SetArgReferee;
 namespace
 {
 const std::string socketName{"/tmp/sessionmanagementservertest-0"};
-constexpr unsigned int socketPermissions{0777};
```
skywilliamsonl, line 44: Not your fault but all these variables should be prefixed with k.

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.cpp
```diff
@@ -182,12 +194,107 @@ void SessionManagementServerTests::serverWillSetLogLevels()
...
+    struct group groupStruct
+    {
+    };
+    groupStruct.gr_gid = kTestGroupId;
+    EXPECT_CALL(*m_linuxWrapperMock, getgrnam_r(testing::StrEq(kTestGroup), _, _, _, _))
+        .WillOnce(testing::DoAll(testing::SetArgPointee<4>(&groupStruct), Return(0)));
+
+    EXPECT_CALL(*m_linuxWrapperMock, chown(testing::StrEq(socketName), kTestOwnerId, kTestGroupId)).WillOnce(Return(0));
+
+    EXPECT_TRUE(m_sut->initialize(socketName, socketPermissions, kTestOwner, kTestGroup));
```
skywilliamsonl, line 227: Instead of having invocation in this function, can you just call sendServerInitialize from the test cases themselves in SessionManagementServerTests.cpp.

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.h
```diff
@@ -55,11 +56,16 @@ class SessionManagementServerTests : public testing::Test
     void sendServerInitialize();
     void sendServerInitializeAndExpectFailure();
     void sendServerStart();
+    void testSocketOwnershipValidOwnerAndGroup();
```
skywilliamsonl, line 59: Can these be renamed to follow the convention in this file, e.g. first one something like:
serverWillInitializeWithValidOwnerAndGroup

Can then move these functions next to the previously definded serverWill/clientWill funcitons.

tests/media/server/ipc/sessionManagementServer/SessionManagementServerTestsFixture.cpp
```diff
@@ -41,11 +43,19 @@ using testing::SetArgReferee;
 namespace
 {
 const std::string socketName{"/tmp/sessionmanagementservertest-0"};
-constexpr unsigned int socketPermissions{0777};
+constexpr unsigned int socketPermissions{0666};
+const std::string socketOwner{};
```
skywilliamsonl, line 47: Maybe call these something like kSocketOwnerEmpty/kSocketGroupEmpty

skywilliamsonl (approved):

PR #208, opened 2023-11-02
Rialto 395 Fix the unit test IpcClientTest_UnexpectedDisconnectWithNotification

skyproudmanp:
Summary: Fix the unit test "IpcClientTest_UnexpectedDisconnectWithNotification" (it intermittently fails)
Type:Fix
Test Plan: Ran unit tests and ran on Sky0 UK hardware 
Jira: RIALTO-395

### DIFF
--- a/media/client/ipc/source/IpcClient.cpp
+++ b/media/client/ipc/source/IpcClient.cpp
@@ -159,11 +159,17 @@ void IpcClient::processIpcThread()
     {
         RIALTO_CLIENT_LOG_ERROR("The ipc channel unexpectedly disconnected, destroying the channel");
 
+        // The test case IpcClientTest.UnexpectedDisconnectWithNotification
+        // will start to destroy the shared_link<> to m_connectionObserver as
+        // soon as the ipcChannel is reset, therefore it's necessary
+        // to create another shared_link<> BEFORE calling m_ipcChannel.reset()
+        // in order to avoid a race since m_connectionObserver is a weak_link...
+        auto connectionObserver{m_connectionObserver.lock()};
+
         // Safe to destroy the ipc objects in the ipc thread as the client has already disconnected.
         // This ensures the channel is destructed and that all ongoing ipc calls are unblocked.
         m_ipcChannel.reset();
 
-        auto connectionObserver{m_connectionObserver.lock()};
         if (connectionObserver)
         {
             connectionObserver->onConnectionBroken();
--- a/media/client/ipc/source/IpcClient.cpp
+++ b/media/client/ipc/source/IpcClient.cpp
@@ -159,17 +159,11 @@ void IpcClient::processIpcThread()
     {
         RIALTO_CLIENT_LOG_ERROR("The ipc channel unexpectedly disconnected, destroying the channel");
 
-        // The test case IpcClientTest.UnexpectedDisconnectWithNotification
-        // will start to destroy the connection observer object
-        // after m_ipcChannel.reset(), therefore it's necessary
-        // to create a shared_ptr<> BEFORE calling m_ipcChannel.reset()
-        // in order to avoid a race (since m_connectionObserver is a weak_ptr)
-        std::shared_ptr<IConnectionObserver> connectionObserver{m_connectionObserver.lock()};
-
         // Safe to destroy the ipc objects in the ipc thread as the client has already disconnected.
         // This ensures the channel is destructed and that all ongoing ipc calls are unblocked.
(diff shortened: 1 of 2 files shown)


media/client/ipc/source/IpcClient.cpp
```diff
@@ -159,11 +159,17 @@ void IpcClient::processIpcThread()
     {
         RIALTO_CLIENT_LOG_ERROR("The ipc channel unexpectedly disconnected, destroying the channel");
 
+        // The test case IpcClientTest.UnexpectedDisconnectWithNotification
```
skywilliamsonl, line 162: This seems like a test issue rather than a source code issue. Can we improve the test case to fix this instead? We would probably need to wait in UnexpectedDisconnectWithNotification for onConnectionBroken to be executed before completing the test.

skyproudmanp: OPTION A
=======

The test is to determine if onConnectionBroken() is called. If the test fails (and the onConnectionBroken is not called) then the test case could wait forever. But perhaps this is OK because it would cause someone to look into the issue and fix it.

OPTION B
=======

Alternatively we could wait a pre-determined amount of time (10 microseconds or some other magic number) for onConnectionBroken to be called. This would be achieved by simply changing the test case like this...

`--- a/tests/media/client/ipc/ipcClient/IpcClientTest.cpp
+++ b/tests/media/client/ipc/ipcClient/IpcClientTest.cpp
@@ -131,6 +131,7 @@ TEST_F(IpcClientTest, UnexpectedDisconnectWithNotification)
     while (m_channelMock.use_count() == ipcChannelCount)
     {
     }
+    usleep(10); // Delay the destruction of the onConnectionBroken() listener

     // On destruction IpcClient does not disconnect
 } `

======

Please let me know which of these options, A or B, seems best

skywilliamsonl: I prefer option A as its deterministic, I would had a timeout to the wait, say 5sec, and fail the test on timeout

tests/media/client/ipc/ipcClient/IpcClientTest.cpp
```diff
@@ -132,5 +134,12 @@ TEST_F(IpcClientTest, UnexpectedDisconnectWithNotification)
     {
     }
 
+    // Wait for the callback
+    constexpr unsigned int kWaitTimeMilliseconds = 5000;
+    for (unsigned int i = 0; !connectionBrokenCallbackCalled && i < kWaitTimeMilliseconds; ++i)
```
skywilliamsonl, line 139: I would expect us to use the wait_for c++ method here along with a condition variable triggered when onConnectionBroken gets called alongstide this bool.

Also, can this test be refactored so that we only have one wait/cond variable? so basically move the m_eventsCond and m_eventsLock to be used here instead.

tests/media/client/ipc/ipcClient/IpcClientTest.cpp
```diff
@@ -99,38 +99,36 @@ TEST_F(IpcClientTest, UnexpectedDisconnectWithNotification)
...
-    // Wait for process to set the ipcChannelCount
+    // Wait for the callback
     {
         std::unique_lock<std::mutex> locker(m_eventsLock);
-        if (!isProcessed)
+        std::chrono::duration kMaximumWaitTime{std::chrono::seconds(5)};
+        if (!connectionBrokenCallbackCalled)
         {
-            m_eventsCond.wait(locker);
+            m_eventsCond.wait_for(locker, kMaximumWaitTime);
```
skywilliamsonl, line 129: Check return status of wait_for, and fail test on timeout:
EXPECT_NE(status, std::cv_status::timeout)

skywilliamsonl (approved):

tests/media/client/ipc/ipcClient/IpcClientTest.cpp
skywojciechowskim: Possible race condition here - connectionBrokenCallbackCalled + notify_all() call can be called between if check and wait_for call. To fix this, you can remove if (!connectionBrokenCallbackCalled) and use wait_for with lambda:
m_eventsCond.wait_for(locker, kMaximumWaitTime, [&](){ return connectionBrokenCallbackCalled; });

skyproudmanp: Doesn't the mutex on lines 116 and 125 prevent a race?

aczs (approved):

PR #212, opened 2023-11-09
Add a unit test for Timer.cpp

skyproudmanp:
Summary: Add a unit test for Timer.cpp
Type: Cleanup
Test Plan: Unit Tests run
Jira: RIALTO-406


tests/common/unittests/main.cpp
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+int main(int argc, char **argv) // NOLINT(build/filename_format)
```
skywilliamsonl, line 22: Because this main function is just the default initialisation of the google tests, we don't need to define this, it is automatically given to use by the gtest framework. We only define main if we need to do some extra init, like for ipc where we do protobuf stuff.

skywilliamsonl (approved):

skywojciechowskim (approved):

PR #216, opened 2023-11-16
ocdm library crash fix

skywojciechowskim:
Summary: Ocdm library crash fix
Type: Fix
Test Plan: FullStack
Jira: RIALTO-355

### DIFF
--- a/media/server/wrappers/source/OcdmSession.cpp
+++ b/media/server/wrappers/source/OcdmSession.cpp
@@ -171,8 +171,9 @@ MediaKeyErrorStatus OcdmSession::storeLicenseData(const uint8_t challenge[], uin
         return MediaKeyErrorStatus::FAIL;
     }
 
-    // Rialto does not support secureStop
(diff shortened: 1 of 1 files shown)


aczs (dismissed):

media/server/wrappers/source/OcdmSession.cpp
skyproudmanp, line 1: I can see why you have 16 bytes. The function opencdm_session_store_license_data() calls StoreLicenseData() which is defined in IOCDM.h (within my soip build) as...

    virtual OCDM_RESULT StoreLicenseData(const uint8_t licenseData[] /* @length:licenseDataSize */,
        uint16_t licenseDataSize,
        uint8_t* secureStopId /* @out @length:16 */)
        = 0;

Perhaps a comment about the origin of this magic number 16 should be included?

skywojciechowskim: replaced with named const.

aczs (approved):

skyproudmanp (approved):

PR #233, opened 2023-12-13
Rialto server media pipeline component tests continued

skywojciechowskim:
Summary: New MediaPipeline Rialto Server component test
Type: Feature
Test Plan: UT, CT
Jira: RIALTO-425


tests/componenttests/server/tests/mediaPipeline/EncryptedPlaybackTest.cpp
```diff
@@ -0,0 +1,342 @@
...
+        EXPECT_CALL(*m_gstWrapperMock, gstByteWriterPutUint16Be(_, segment->getSubSamples().front().numClearBytes));
+        EXPECT_CALL(*m_gstWrapperMock, gstByteWriterPutUint32Be(_, segment->getSubSamples().front().numEncryptedBytes));
+        EXPECT_CALL(*m_gstWrapperMock, gstBufferNewWrapped(&subsamples, kSubsamplesSize)).WillOnce(Return(&subsamplesBuffer));
+        EXPECT_CALL(*m_gstWrapperMock, gstBufferAddMeta(&dataBuffer, _, _)).WillOnce(Return(&meta));
+    }
+
+    void pushEncryptedAudioData(int needDataFrameCount)
+    {
+        // First, generate new data
+        auto segment{SegmentBuilder().basicAudioSegment(m_audioSourceId).withEncryptionData()()};
```
skywilliamsonl, line 75: Do you think the encrypted data written to segments should be associated with a mediaKeySession instead of been random data? Not something to do now but maybe an improvement in the future to make this test more realistic.

skywilliamsonl (approved):

skyproudmanp (approved):

tests/componenttests/server/tests/mediaPipeline/EncryptedPlaybackTest.cpp
skywojciechowskim: We don't have MediaKeys tests yet, but we can think about it. But it may be hard to combine MediaPipeline and MediaKeys tests in Rialto Server, because Gstreamer is wrapped and GstRialtoDecryptor element is not created

PR #240, opened 2023-12-27
Fix socket leak in component tests

skywilliamsonl:
Summary: Close server socket in server component tests.
Type: Fix
Test Plan: Unittests, Component Tests.
Jira: NO-JIRA


skyproudmanp (approved):
Should we test the return values from shutdown() and close() ?

aczs (approved):

PR #246, opened 2024-01-08
Rialto 427 component test server media keys

skyproudmanp:
Summary: Component tests for server media keys
Type: Feature
Test Plan: Unit Tests and other component tests
Jira: RAILTO-427

### DIFF
--- a/tests/componenttests/server/tests/mediaKeys/SessionReadyForDecryptionTest.cpp
+++ b/tests/componenttests/server/tests/mediaKeys/SessionReadyForDecryptionTest.cpp
@@ -1,0 +1,209 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2024 Sky UK
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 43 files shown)


skywojciechowskim (approved):

tests/componenttests/server/tests/mediaKeys/SessionReadyForDecryptionTest.cpp
```diff
@@ -0,0 +1,264 @@
...
+ * Code:
+ */
+TEST_F(SessionReadyForDecryptionTest, shouldUpdateWidevine)
+{
+    createMediaKeysWidevine();
+    ocdmSessionWillBeCreated();
+    createKeySession();
+
+    // Step 1: Generate license request
+    licenseRenew();
```
skywilliamsonl, line 174: I would expect a call to generate request with a license request callback rather than renewal here.

Additionally, in the tests in client side the generate request step is subtlety different for both widevine and netflix, for Netflix/playready we expect a license request notification before the generateRequest API call returns, for widevine generateRequest completes and then we receive a notification. This can be seen in the Create Cobalt and Create Netflix sequence diagrams.

skywojciechowskim (approved):

skywilliamsonl (approved):

PR #270, opened 2024-02-15
Removed unneeded environment variables

aczs:
Summary: Removed unneeded environment variables
Type: Feature
Test Plan: Fullstack, Unittests, Component Tests.
Jira: RIALTO-454

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,7 +137,7 @@ option(ENABLE_SERVER "Enable building RialtoServer" ON)
 option(ENABLE_SERVER_MANAGER "Enable building RialtoServerManagerSim" ON)
 
 if ( NOT ENVIRONMENT_VARIABLES)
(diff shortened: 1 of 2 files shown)


skyproudmanp (commented):
Should WAYLAND_DISPLAY also be removed from:-

rialto/tests/unittests/media/server/service/playbackService/PlaybackServiceTestsFixture.cpp lines 153 and 154

aczs:
> Should WAYLAND_DISPLAY also be removed from:-
> 
> rialto/tests/unittests/media/server/service/playbackService/PlaybackServiceTestsFixture.cpp lines 153 and 154

No. WAYLAND_DISPLAY env variable still must be set in RialtoServer, because westerossink needs it.
Test mentioned by you tests scenario when display name is set by RialtoServerManager and that's still valid scenario.

skyshaha1 (approved):

PR #302, opened 2024-05-20
Fix some problems reported by valgrind and helgrind

skyproudmanp:
Summary: Fix some problems reported by valgrind and helgrind on the native build
Type: Fix
Test Plan: UT/ CT
Jira: NO-JIRA

### DIFF
--- a/media/server/ipc/include/SetLogLevelsService.h
+++ b/media/server/ipc/include/SetLogLevelsService.h
@@ -20,10 +20,12 @@
 #ifndef FIREBOLT_RIALTO_SERVER_IPC_SET_LOG_LEVELS_SERVICE_H_
 #define FIREBOLT_RIALTO_SERVER_IPC_SET_LOG_LEVELS_SERVICE_H_
 
-#include "RialtoServerLogging.h"
-#include <IIpcServer.h>
 #include <memory>
+#include <mutex>
 #include <set>
+
+#include "IIpcServer.h"
+#include "RialtoServerLogging.h"
 
 namespace firebolt::rialto::server::ipc
 {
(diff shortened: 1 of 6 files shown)


media/server/service/source/SessionServerManager.cpp
```diff
@@ -226,6 +228,7 @@ bool SessionServerManager::switchToInactive()
 
 bool SessionServerManager::switchToNotRunning()
 {
+    std::unique_lock<std::mutex> lock{m_switchStateMutex};
```
aczs, line 231: What's the problem here? That order of sending sendStateChangedEvent might be incorrect in extremely rare case? If I'm correct:
1. Have you managed to reproduce it without adding extra sleeps?
2. Why the mutex protects only switching to NotRunning? Other state changes can be also affected by this race, right?
3. What do you think about moving sendStateChangedEvent(common::SessionServerState::UNINITIALIZED); into ipc thread (into lambda in ApplicationManagementServer::start())? Wouldn't that synchronise the flow without mutexes?

media/server/ipc/include/SetLogLevelsService.h
```diff
@@ -20,11 +20,13 @@
 #ifndef FIREBOLT_RIALTO_SERVER_IPC_SET_LOG_LEVELS_SERVICE_H_
 #define FIREBOLT_RIALTO_SERVER_IPC_SET_LOG_LEVELS_SERVICE_H_
 
-#include "RialtoServerLogging.h"
-#include <IIpcServer.h>
 #include <memory>
+#include <mutex>
 #include <set>
 
+#include "IIpcServer.h"
```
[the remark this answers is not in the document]
skyproudmanp: All headers are required except "RialtoServerLogging.h" can be replaced by "RialtoLogging.h" since the code beneath just requires the definition of RIALTO_DEBUG_LEVEL

media/server/service/source/SessionServerManager.cpp
```diff
@@ -235,12 +245,21 @@ bool SessionServerManager::switchToNotRunning()
     m_playbackService.switchToInactive();
     m_cdmService.switchToInactive();
     m_controlService.setApplicationState(ApplicationState::UNKNOWN);
-    stopService();
+
+    bool rv{true};
```
aczs, line 249: Please rename rv to result.

aczs (approved):

PR #303, opened 2024-05-22
Valgrind on native

skyproudmanp:
Summary: Write test infrastructure for testing rialto with valgrind on the native platform
Type: Feature
Test Plan: UT/ CT  
Jira: RIALTO-258


CMakeLists.txt
```diff
@@ -206,6 +207,8 @@ endif()
 if( NATIVE_BUILD )
     add_subdirectory( stubs/rdk_gstreamer_utils )
     add_subdirectory( stubs/opencdm )
+    add_compile_options(-DFREE_MEM_BEFORE_EXIT)
+    add_compile_options(-O0 -ggdb)
```
aczs, line 211: please add an additional flag for debugging with GDB

media/server/service/source/main.cpp
```diff
@@ -47,13 +51,32 @@ int main(int argc, char *argv[])
...
+            firebolt::rialto::server::gstDeinitalise();
+            return EXIT_FAILURE;
+        }
+
+        appSessionServer->startService();
+    }
 
-    if (!appSessionServer->init(argc, argv))
+#ifdef FREE_MEM_BEFORE_EXIT
+    if (getenv("RIALTO_FREE_MEM_BEFORE_EXIT"))
```
aczs, line 71: Do we need both FREE_MEM_BEFORE_EXIT compiler flag and the env var? I would stay with the flag only.

PR #306, opened 2024-06-12
Register and unregister clients with proxy to fix data race, and tidy the surrounding code

skyproudmanp:
Summary: register and unregister clients (of control) via a proxy (this prevents a data race and updates c style pointers to c++). The call to registerClient in ControlService::addControl was deceiving since this didn't actually register any client... it just called setApplicationState therefore changed to call setApplicationState() direct
Type: Fix
Test Plan: UT/CT and test webaudio on XiOne US box
Jira: NO-JIRA


jamcoole:
sanity test failures are most likely unrelated to this change, there are some regressions on rdk-next for RPI currently being investigated, have re-triggered and will up-vote if it fails again

media/client/main/source/ClientController.cpp
```diff
@@ -102,26 +111,54 @@ bool ClientController::registerClient(IControlClient *client, ApplicationState &
         }
     }
     m_registrationRequired = false;
-    m_clientVec.insert(client);
+
+    bool alreadyRegistered{false};
+    for (auto i = m_clients.begin(); i != m_clients.end(); ++i)
```
skywojciechowskim, line 116: std find_if :)

media/public/include/IControl.h
```diff
@@ -81,14 +86,18 @@ class IControl
...
+     *        At destruction the client will be unregistered and the
+     *        shared_ptr will be released
      *
      * @param[in]  client   : Client object for callbacks
      * @param[out] appState : Current application state
      *
      * @retval true on success, false otherwise.
      */
-    virtual bool registerClient(std::weak_ptr<IControlClient> client, ApplicationState &appState) = 0;
+    virtual bool registerClientAndUnregisterOnDestruction(std::weak_ptr<IControlClient> client,
```
skywojciechowskim, line 99: If you want to change method name only, I'd keep the old one to skip public interface change

media/server/main/interface/IControlServerInternal.h
```diff
@@ -61,11 +61,11 @@ class IControlServerInternalFactory : public IControlFactory
     createControlServerInternal(int id, const std::shared_ptr<IControlClientServerInternal> &client) const = 0;
 };
 
-class IControlServerInternal : public IControl
```
skywojciechowskim, line 64: All ServerInternal objects inherit from public ones. Please keep it. What if we add a new method to IControl in future?

media/public/include/IWebAudioPlayer.h
```diff
@@ -29,12 +29,13 @@
  * current audio output.
  */
 
-#include "IWebAudioPlayerClient.h"
```
skywojciechowskim, line 32: If you want to change include order only, I'd keep the old one to avoid public interface change

media/public/include/IControl.h
skyproudmanp: But the old name didn't say what he method actually does...

media/public/include/IWebAudioPlayer.h
skyproudmanp: Does this actually change the interface? And the update conforms to the google coding standards

media/public/include/IControl.h
skywojciechowskim: ok, so let's wait for Adam and Luke opinion about that ;)

media/server/main/interface/IControlServerInternal.h
skyproudmanp: Then perhaps it should inherit from something else? Perhaps a public interface class that has no methods?
I think that it's wrong that this inherits from IControl when nothing actually needs ServerInternal to implement any of the methods that are required by the IControl interface?

But... if you have plans for something in the future then I'll leave it..

media/public/include/IWebAudioPlayer.h
skywojciechowskim: ok, so you can change that

media/server/main/interface/IControlServerInternal.h
skywojciechowskim: That's why we keep that inheritance - to be ready for public interface changes, which have to be handled both on server and client side ;)

media/client/main/source/MediaPipeline.cpp
```diff
@@ -134,15 +136,16 @@ MediaPipelineFactory::createMediaPipeline(std::weak_ptr<IMediaPipelineClient> cl
...
-                                                                    ? *clientControllerLocked
-                                                                    : client::IClientControllerAccessor::instance()
-                                                                          .getClientController());
+        firebolt::rialto::client::IClientController &cc =
+            clientControllerLocked ? *clientControllerLocked
+                                   : client::IClientControllerAccessor::instance().getClientController();
+
+        auto mp{std::make_shared<client::MediaPipeline>(client, videoRequirements,
+                                                        mediaPipelineIpcFactoryLocked
+                                                            ? mediaPipelineIpcFactoryLocked
```
skywilliamsonl, line 145: What about a new registerclient api that's called from constructor?

skywojciechowskim: I discussed that with Phil earlier. The problem is that unregister client is called in destructor and (with enable_shared_from_this), shared_from_this doesn't work in destructor (makes sense, it's a virtual function). I suggested to call unregister client in shared_ptr custom deleter, but it also throws an exception

media/public/include/IControl.h
skyproudmanp: Name is reverted back to registerClient and a comment is used to highlight the difference between IClientController::registerClient and IControl::registerClient

media/public/include/IControl.h
```diff
@@ -89,15 +84,16 @@ class IControl
      * @brief Register new IControlClient. This method will hold a shared_ptr
      *        to the client until the destruction of the IControl object.
      *        At destruction the client will be unregistered and the
-     *        shared_ptr will be released
+     *        shared_ptr will be released. NOTE: this is different
```
skywilliamsonl, line 87: I dont think this comment should be in the public interface, IClientController isn't a public interface so the user wont necessarily know what it is.

tests/unittests/media/client/main/mediaPipeline/MediaPipelineProxyTest.cpp
```diff
@@ -0,0 +1,142 @@
...
+    EXPECT_CALL(*mediaPipelineMock, getMute(_)).WillOnce(DoAll(SetArgReferee<0>(false), Return(true)));
+
+    EXPECT_CALL(*mediaPipelineMock, flush(kSourceId, true)).WillOnce(Return(true));
+    EXPECT_CALL(*mediaPipelineMock, setSourcePosition(kSourceId, kPosition1)).WillOnce(Return(true));
+    EXPECT_CALL(*mediaPipelineMock, getClient()).WillOnce(Return(kIMediaPipelineClient));
+    EXPECT_CALL(*mediaPipelineMock, notifyApplicationState(ApplicationState::RUNNING));
+
+    /////////////////////////////////////////////
+
+    EXPECT_TRUE(proxy->load(MediaType::MSE, kMimeType, kUrl));
```
skywilliamsonl, line 101: I would prefer the EXPECT_CALL to be set just before we call the method, e.g.:

EXPECT_CALL(*mediaPipelineMock, attachSource(_)).WillOnce(Return(true));
EXPECT_TRUE(proxy->attachSource(kMediaSource));

For example what is proxy->play() actually called media_pipeline->pause() for some strange reason. EXPECT_TRUE(proxy->play()); would still succeed because we have set EXPECT_CALL(*mediaPipelineMock, pause()).WillOnce(Return(true)); prior to this. EXPECT_TRUE(proxy->pause()); would then fail even though it behaved correctly.

skywilliamsonl (approved):

skywojciechowskim (approved):

PR #307, opened 2024-07-11
Flush thread events between component tests

skyproudmanp:
Summary: The pull request 306 highlighted an intermittent problem in component tests that allowed the teardown of one test to still be running while the next test starts. This was potentially happening in different threads but for certain it happened within ControlIpc's thread. Extra flush calls in this PR will wait for all threads to finish before teardown is allowed to complete. This wasn't a problem before 306 because the events concerned didn't hold a lock on the WebAudioPlayer/ MediaPipeline clients which allowed them to start destructing while they were still potentially being used (therefore destruction used to happen earlier). Now the destruction is correctly delayed until the final thread drops its shared_ptr (and the next test now also delayed until immediately after this happens).    
Type: Fix  
Test Plan: UT/ CT  
Jira: NO-JIRA

### DIFF
--- a/media/client/ipc/include/ControlIpc.h
+++ b/media/client/ipc/include/ControlIpc.h
@@ -71,6 +71,7 @@ public:
 
     bool getSharedMemory(int32_t &fd, uint32_t &size) override;
     bool registerClient() override;
+    void eventThreadFlush() override;
 
 private:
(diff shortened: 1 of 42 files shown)


skyproudmanp:
Luke suggested an easier way to prevent this problem

PR #314, opened 2024-08-03
Rebase 0001-no-vp9-on-rpi.patch

HarshiniRavichandran:
[RDKCMF-8884] Update reference image for new rialto: fix RPI
DAC image enables vp9 decoder, but vp9 does not work on RPI. Disable it in rialto on RPI host.
  

[ RDKCMF-8899] reference image: bump rialto version
A newer rialto version is needed to be compatible with newer DAC apps.

### DIFF
--- a/media/server/gstplayer/include/GstMimeMapping.h
+++ b/media/server/gstplayer/include/GstMimeMapping.h
@@ -48,7 +48,7 @@ inline GstCaps *createSimpleCapsFromMimeType(std::shared_ptr<firebolt::rialto::w
 {
     static const std::unordered_map<std::string, std::string> mimeToMediaType =
         {{"video/h264", "video/x-h264"}, {"video/h265", "video/x-h265"},   {"video/x-av1", "video/x-av1"},
-         {"video/x-vp9", "video/x-vp9"}, {"video/mp4", "video/mpeg"},      {"audio/mp4", "audio/mpeg"},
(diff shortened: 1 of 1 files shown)


media/server/gstplayer/include/GstMimeMapping.h
```diff
@@ -81,12 +81,10 @@ convertFromCapsVectorToMimeSet(const std::vector<GstCaps *> &supportedCaps,
          {m_gstWrapper->gstCapsFromString("video/x-av1"), {"video/x-av1"}},
          {m_gstWrapper->gstCapsFromString("video/x-h264"), {"video/h264"}},
          {m_gstWrapper->gstCapsFromString("video/x-h265"), {"video/h265"}},
-         {m_gstWrapper->gstCapsFromString("video/x-vp9"), {"video/x-vp9"}},
```
aczs, line 84: This change will disable VP9 on all platforms.
Additionally, Rialto will only return that VP9 is supported if one of the platform's GST decoders or sinks return it is supported. 
If GStreamer says VP9 is supported but it doesn't actually work, it is not Rialto's problem and workarounds for that problem shouldn't be applied in Rialto.
If there's no other way, I would suggest to use bitbake patches for the affected platform only.

aczs (changes requested):

svenko766 (changes requested):
This PR is not required as rebasing the patch file should work.  The patch is applied only for Rpi platforms. So, it wont affect the other platforms.  Please abandon the PR

PR #322, opened 2024-08-23
Position reporting after setSourcePosition fix for amlogic

skywojciechowskim:
Summary: Position reporting after setSourcePosition fix for amlogic
Type: Fix
Test Plan: UT, Fullstack
Jira: RIALTO-620

### DIFF
--- a/media/client/ipc/include/MediaPipelineIpc.h
+++ b/media/client/ipc/include/MediaPipelineIpc.h
@@ -107,7 +107,7 @@ public:
 
     bool flush(int32_t sourceId, bool resetTime) override;
 
-    bool setSourcePosition(int32_t sourceId, int64_t position) override;
+    bool setSourcePosition(int32_t sourceId, int64_t position, bool resetTime) override;
 
     bool processAudioGap(int64_t position, uint32_t duration, int64_t discontinuityGap, bool audioAac) override;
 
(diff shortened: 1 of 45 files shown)


media/server/gstplayer/include/GenericPlayerContext.h
```diff
@@ -46,6 +47,15 @@ struct Rectangle
     inline void clear() { x = y = width = height = 0; }
 };
 
+/**
+ * @brief Structure used for set source position
+ */
+struct SegmentData
+{
+    int64_t position;
```
skyproudmanp, line 55: Should this be uint64_t ?

I might have spotted something :-)

media/client/ipc/include/MediaPipelineIpc.h
```diff
@@ -107,7 +107,7 @@ class MediaPipelineIpc : public IMediaPipelineIpc, public IpcModule
 
     bool flush(int32_t sourceId, bool resetTime) override;
 
-    bool setSourcePosition(int32_t sourceId, int64_t position) override;
+    bool setSourcePosition(int32_t sourceId, int64_t position, bool resetTime) override;
```
skyproudmanp, line 110: Again, should position perhaps be uint64_t because the call to gstSegmentDoSeek is unsigned?

media/server/gstplayer/include/GenericPlayerContext.h
skywojciechowskim: no, position has to be int64_t

skyproudmanp (approved):

skyshaha1 (approved):

PR #328, opened 2024-09-05
Fix getSupportedProperties method so that it can reliably determine if immediate-output, etc, is supported

skyproudmanp:
Summary: fix_getSupportedProperties so that it is robust. For speed, initially, it tries to get properties from the class info only. If this fails, and it often does fail immediately after rialtoserver start-up, then it uses gst_plugin_feature_load and tries again. If this fails then it instantiates an object in order to get its properties (this is what gst-inspect does)
Type: Fix  
Test Plan: UT, CT, native_build
Jira: RIALTO-597

### DIFF
--- a/media/server/gstplayer/source/GstCapabilities.cpp
+++ b/media/server/gstplayer/source/GstCapabilities.cpp
@@ -207,28 +207,102 @@ std::vector<std::string> GstCapabilities::getSupportedProperties(MediaSourceType
     for (GList *iter = factories; iter != nullptr && !propertiesToLookFor.empty(); iter = iter->next)
     {
         GstElementFactory *factory = GST_ELEMENT_FACTORY(iter->data);
-        GType elementType = m_gstWrapper->gstElementFactoryGetElementType(factory);
-        if (elementType == G_TYPE_INVALID)
-            continue;
-        gpointer elementClass = m_glibWrapper->gTypeClassRef(elementType);
-        if (elementClass)
+        gpointer elementClass{nullptr};
+        GstPluginFeature *loadFeature{nullptr};
+        GstElement *elementObj{nullptr};
+
+        RIALTO_SERVER_LOG_DEBUG("Searching the properties of a class");
         {
-            for (auto i = propertiesToLookFor.begin(); i != propertiesToLookFor.end();)
+            // Try to obtain the class without instantiating an object...
+            GType elementType = m_gstWrapper->gstElementFactoryGetElementType(factory);
+            if (elementType == G_TYPE_INVALID)
             {
-                if (m_glibWrapper->gObjectClassFindProperty(G_OBJECT_CLASS(elementClass), i->c_str()))
-                {
-                    propertiesFound.push_back(*i);
-                    i = propertiesToLookFor.erase(i);
-                }
-                else
-                    ++i;
+                RIALTO_SERVER_LOG_DEBUG("  Using gstPluginFeatureLoad");
+                loadFeature = m_gstWrapper->gstPluginFeatureLoad(GST_PLUGIN_FEATURE(factory));
+                if (loadFeature)
+                    elementType = m_gstWrapper->gstElementFactoryGetElementType(factory);
             }
 
-            m_glibWrapper->gObjectUnref(elementClass);
+            if (elementType != G_TYPE_INVALID)
+                elementClass = m_glibWrapper->gTypeClassRef(elementType);
         }
+
+        if (!elementClass)
+        {
+            // Create an object because we couldn't get the class
+            RIALTO_SERVER_LOG_DEBUG("  Using gstElementFactoryCreate");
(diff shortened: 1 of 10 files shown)


media/server/gstplayer/source/GstCapabilities.cpp
```diff
@@ -207,28 +207,102 @@ std::vector<std::string> GstCapabilities::getSupportedProperties(MediaSourceType
...
-        if (elementClass)
+        gpointer elementClass{nullptr};
+        GstPluginFeature *loadFeature{nullptr};
+        GstElement *elementObj{nullptr};
+
+        RIALTO_SERVER_LOG_DEBUG("Searching the properties of a class");
         {
-            for (auto i = propertiesToLookFor.begin(); i != propertiesToLookFor.end();)
+            // Try to obtain the class without instantiating an object...
+            GType elementType = m_gstWrapper->gstElementFactoryGetElementType(factory);
```
skyshaha1, line 217: This section of logic to handle the getting of the element type is duplicated quite a bit, maybe think about creating a separate function that would have that logic to avoid repetition and make the code cleaner

media/server/gstplayer/source/GstCapabilities.cpp
```diff
@@ -207,28 +207,102 @@ std::vector<std::string> GstCapabilities::getSupportedProperties(MediaSourceType
...
+                // Fall back to create an object
+                RIALTO_SERVER_LOG_DEBUG("  No properties seen, creating an object");
+                elementObj = m_gstWrapper->gstElementFactoryCreate(factory, nullptr);
+                if (elementObj)
+                    props = m_glibWrapper->gObjectClassListProperties(G_OBJECT_GET_CLASS(elementObj), &nProps);
+            }
+
+            if (props)
+            {
+                for (guint j = 0; j < nProps && !propertiesToLookFor.empty(); ++j)
```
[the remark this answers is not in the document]
skyproudmanp: Nice suggestion!

media/server/gstplayer/source/GstCapabilities.cpp
skyproudmanp: Are you talking about these three lines that occur twice (once with XXX blank and once where XXX contains some additional code)...

                GType elementType = m_gstWrapper->gstElementFactoryGetElementType(factory);
                XXX
                if (elementType != G_TYPE_INVALID)
                    elementClass = m_glibWrapper->gTypeClassRef(elementType);
...you're thinking there should be a function for the last two lines?

skyshaha1: Yeh so some sort of helper function, I was thinking one to retrieve the element class, getElementClass, and another helper function to fetch properties from either the class or the object, getElementProperties maybe?

aczs (approved):

tests/componenttests/server/tests/mediaPipelineCapabilities/MediaPipelineCapabilitiesTest.cpp
```diff
@@ -37,23 +46,32 @@ class MediaPipelineCapabilitiesTest : public RialtoServerComponentTest
...
             GST_ELEMENT_FACTORY_TYPE_SINK | GST_ELEMENT_FACTORY_TYPE_DECODER | GST_ELEMENT_FACTORY_TYPE_MEDIA_VIDEO};
         m_listOfFactories = g_list_append(m_listOfFactories, dummyFactory);
         EXPECT_CALL(*m_gstWrapperMock, gstElementFactoryListGetElements(expectedFactoryListType, GST_RANK_NONE))
             .WillOnce(Return(m_listOfFactories));
-        EXPECT_CALL(*m_gstWrapperMock, gstElementFactoryGetElementType(dummyFactory)).WillOnce(Return(kDummyType));
-        EXPECT_CALL(*m_glibWrapperMock, gTypeClassRef(kDummyType)).WillOnce(Return(&m_dummyClass));
-        EXPECT_CALL(*m_glibWrapperMock, gObjectClassFindProperty(&m_dummyClass, _)).WillRepeatedly(Return(&m_dummyParam));
-        EXPECT_CALL(*m_glibWrapperMock, gObjectUnref(&m_dummyClass));
+        // The next calls should ensure that an object is created and then freed
+        GstElement object;
```
skyfroxl, line 67: Variables here should be in the test class/global, EXPECT_CALLs are initialised here but aren't actually 'used' until the mock is called in the test. When the mock is called, these objects will be out of scope which can cause problems with the test.

skyproudmanp: Actually, I'm wrong... G_OBJECT_GET_CLASS() is called on &object and therefore it's an excellent point that you're making! (And that's why I had to do the memset in the above code). I'm moving it...

skyfroxl (approved):

aczs (approved):

PR #329, opened 2024-09-11
Require Audio Easing support on RialtoMSEAudioSink for Netflix (audio-fade and fade-volume props)

skyshaha1:
Summary: Require Audio Easing support on RialtoMSEAudioSink for Netflix
Type: Feature
Test Plan: UT/ CT
Jira: RIALTO-593


media/server/gstplayer/source/GstCapabilities.cpp
```diff
@@ -233,6 +245,10 @@ std::vector<std::string> GstCapabilities::getSupportedProperties(MediaSourceType
                         RIALTO_SERVER_LOG_DEBUG("Found property '%s'", kPropName.c_str());
                         propertiesFound.push_back(kPropName);
                         propertiesToLookFor.erase(it);
+                        if(kPropName == "audio-fade")
```
skyfroxl, line 248: Add a comment to explain this

media/server/gstplayer/source/tasks/generic/SetVolume.cpp
```diff
@@ -35,15 +43,74 @@ SetVolume::~SetVolume()
...
+        return firebolt::rialto::wrappers::rgu_Ease::EaseLinear;
+    case EaseType::EASE_IN_CUBIC:
+        return firebolt::rialto::wrappers::rgu_Ease::EaseInCubic;
+    case EaseType::EASE_OUT_CUBIC:
+        return firebolt::rialto::wrappers::rgu_Ease::EaseOutCubic;
+    default:
+        throw std::invalid_argument("Unknown EaseType");
+    }
+
+    return firebolt::rialto::wrappers::rgu_Ease::EaseLinear;
```
aczs, line 60: this return will never be invoked

media/server/gstplayer/source/tasks/generic/SetVolume.cpp
```diff
@@ -35,15 +43,74 @@ SetVolume::~SetVolume()
...
     if (!m_context.pipeline)
     {
         RIALTO_SERVER_LOG_ERROR("Setting volume failed. Pipeline is NULL");
         return;
     }
-    m_gstWrapper->gstStreamVolumeSetVolume(GST_STREAM_VOLUME(m_context.pipeline), GST_STREAM_VOLUME_FORMAT_LINEAR,
-                                           m_volume);
+    GstElement *audioSink = m_player.getSink(firebolt::rialto::MediaSourceType::AUDIO);
+
+    if (audioSink && m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(audioSink), "audio-fade"))
```
aczs, line 74: Can you explain how normal set volume request is going to work? I suppose it will try to use this task, right? What will happen if the client will request normal change of volume from let's say 0.5 to 1.0? Is it going to use fade-volume property too?

skyshaha1: Yes so a normal set volume request will use the SetVolume task. So when the vol would change from 0.5 to 1.0, it would trigger the SetVol task and the constructor would take the target volume to be 1.0 (in this case), as well as the volDuration and easingType. So then if the sink exists and supports the audio-fade prop, then it will be used for transition, from 0.5 to 1.0. If the audio-fade prop is not available, it will check if the SoC supports audio-fade and apply the easing using doAudioEasingonSoc(). If neither are supported then it will use gstStreamVolumeSetVolume() without any audio fade effect. So looking at the fade-volume property in the netflix code, it is used in getFadeVolume, and in our case it would get the current volume, so yeh I think it would use the fade-volume property, but I can double check on that

aczs: And why we want to use audio fade for the normal volume setting? Normal volume setting should apply the change immediately, not in gradual way as "audio-fade" does.

aczs (changes requested):

media/server/gstplayer/source/tasks/generic/SetVolume.cpp
aczs: Besides, is incremental audio fade actually a thing?

skyshaha1: Oh ok so we basically would want to skip the use of audio-fade in a normal volume change. So I can adjust the logic to only apply audio-fade when a fade-duration is explicitly requested, by including a flag to check that if the m_volumeDuration is 0, then it does an immediate change, otherwise it does volume transition using audio fade

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -408,6 +408,14 @@ GstElement *GstGenericPlayer::getSink(const MediaSourceType &mediaSourceType)
     }
     else
     {
+        if (m_context.pipeline == nullptr)
```
skyfroxl, line 411: These logs needed? Phil is doing some refactoring of getSink so I would leave as is for now.

wrappers/source/RdkGstreamerUtilsWrapper.cpp
```diff
@@ -52,4 +53,31 @@ void RdkGstreamerUtilsWrapper::processAudioGap(GstElement *pipeline, gint64 gaps
...
+        return rdk_gstreamer_utils::rgu_Ease::EaseOutCubic;
+    default:
+        throw std::invalid_argument("Unknown EaseType");
+    }
+}
+
+void RdkGstreamerUtilsWrapper::doAudioEasingonSoc(double target, uint32_t duration, rgu_Ease ease) const
+{
+    rdk_gstreamer_utils::rgu_Ease convertedEaseType =
+        convertEaseType(static_cast<firebolt::rialto::wrappers::EaseType>(ease));
```
skyfroxl, line 75: Do you need a conversion here?

tests/componenttests/server/tests/mediaPipelineCapabilities/MediaPipelineCapabilitiesTest.cpp
```diff
@@ -35,6 +35,7 @@ constexpr int kNumPropertiesOnSink{3};
 const char *kPropertyName1 = "test-name-5";
 const char *kPropertyName2 = "test2";
 const char *kPropertyName3 = "prop";
+const char *kPropertyName4 = "audio-fade";
```
skyfroxl, line 38: I would name this constant kAudioFadePropertyName. The others are generic but this one is not.

tests/unittests/media/client/main/mediaPipeline/SetVolumeTest.cpp
```diff
@@ -44,17 +46,17 @@ class RialtoClientMediaPipelineSetVolumeTest : public MediaPipelineTestBase
  */
 TEST_F(RialtoClientMediaPipelineSetVolumeTest, setVolumeSuccess)
 {
-    EXPECT_CALL(*m_mediaPipelineIpcMock, setVolume(m_kVolume)).WillOnce(Return(true));
```
skyfroxl, line 47: Add a test to test the API when only volume is passed it so easetype and duration should be default.

tests/unittests/media/server/gstplayer/genericPlayer/GstCapabilitiesTest.cpp
```diff
@@ -92,7 +95,17 @@ class GstCapabilitiesTest : public testing::Test
     {
         IFactoryAccessor::instance().gstWrapperFactory() = nullptr;
         IFactoryAccessor::instance().glibWrapperFactory() = nullptr;
-        gst_object_unref(m_elementFactory);
+        IFactoryAccessor::instance().rdkGstreamerUtilsWrapperFactory() = nullptr;
+        if (m_elementFactory)
+        {
+            GstObject *parent = gst_object_get_parent(GST_OBJECT(m_elementFactory));
```
skyfroxl, line 101: This getting and unreffing the parent is redundant code. We shouldn't need to do it.

tests/unittests/media/server/gstplayer/genericPlayer/GstCapabilitiesTest.cpp
```diff
@@ -340,9 +359,10 @@ TEST_F(GstCapabilitiesTest, getSupportedPropertiesWithPropertiesSupported)
     EXPECT_CALL(*m_glibWrapperMock, gObjectClassListProperties(_, _))
         .WillOnce(DoAll(SetArgPointee<1>(kNumParamsSupportedByServer), Return(dummySinkParamsPtr)));
     EXPECT_CALL(*m_glibWrapperMock, gFree(dummySinkParamsPtr)).Times(1);
+    EXPECT_CALL(*m_rdkGstreamerUtilsWrapperMock, isSocAudioFadeSupported()).WillOnce(Return(true));
```
skyfroxl, line 362: I would keep this test as it is, and dont pass 'audio-fade' in, in this case isSocAudioFadeSupported shouldnt be called.

Then you need two separate tests, one for audio-fade property and one for soc.

tests/unittests/media/server/gstplayer/genericPlayer/GstCapabilitiesTest.cpp
```diff
@@ -365,12 +385,14 @@ TEST_F(GstCapabilitiesTest, getSupportedPropertiesWithNoPropertiesSupported)
...
-
+    EXPECT_CALL(*m_rdkGstreamerUtilsWrapperMock, isSocAudioFadeSupported()).WillOnce(Return(true));
     // Params that the caller is asking about...
     std::vector<std::string> kParamNames{"test-name-123", "test2"};
     std::vector<std::string> supportedProperties{m_sut->getSupportedProperties(MediaSourceType::VIDEO, kParamNames)};
     // this time we will not find the properties...
-    EXPECT_TRUE(supportedProperties.empty());
+    // isAudioSupported is initialised as false, as it doesn't go into the if statement of the element class it remains
+    // false - which is expected
+    EXPECT_EQ(supportedProperties, std::vector<std::string>{"audio-fade"});
```
skyfroxl, line 395: This test should return no properties and should call isSocAudioFadeSupported.

tests/unittests/media/server/gstplayer/genericPlayer/common/GenericTasksTestsBase.cpp
```diff
@@ -777,18 +781,157 @@ void GenericTasksTestsBase::triggerSetupSource()
...
+    switch (easeType)
+    {
+    case EaseType::EASE_LINEAR:
+        return firebolt::rialto::wrappers::rgu_Ease::EaseLinear;
+    case EaseType::EASE_IN_CUBIC:
+        return firebolt::rialto::wrappers::rgu_Ease::EaseInCubic;
+    case EaseType::EASE_OUT_CUBIC:
+        return firebolt::rialto::wrappers::rgu_Ease::EaseOutCubic;
+    default:
+        throw std::invalid_argument("Invalid EaseType provided");
```
skyfroxl, line 859: Better to use a test assert.

media/server/gstplayer/source/GstGenericPlayer.cpp
skyshaha1: Yeh these logs aren't needed at all

tests/componenttests/server/tests/mediaPipeline/VolumeTest.cpp
```diff
@@ -51,12 +51,24 @@ class VolumeTest : public MediaPipelineTest
...
                 }));
 
-        EXPECT_CALL(*m_glibWrapperMock, gObjectClassFindProperty(_, StrEq("audio-fade"))).WillOnce(Return(nullptr));
-
-        EXPECT_CALL(*m_rdkGstreamerUtilsWrapperMock, isSocAudioFadeSupported()).WillOnce(Return(false));
-
+        // Normal volume change when volumeDuration == 0
+        if (m_volumeDuration == 0)
+        {
+            EXPECT_CALL(*m_rdkGstreamerUtilsWrapperMock, isSocAudioFadeSupported()).Times(0);
```
skyfroxl, line 57: Remove expect calls that require .Times(0), these are strict mocks so it is enforced automatically.

tests/unittests/media/server/gstplayer/genericPlayer/common/GenericTasksTestsBase.cpp
```diff
@@ -856,7 +856,8 @@ firebolt::rialto::wrappers::rgu_Ease convertEaseType(firebolt::rialto::EaseType
     case EaseType::EASE_OUT_CUBIC:
         return firebolt::rialto::wrappers::rgu_Ease::EaseOutCubic;
     default:
-        throw std::invalid_argument("Invalid EaseType provided");
+        ADD_FAILURE() << "Invalid EaseType provided: " << static_cast<int>(easeType);
```
skyfroxl, line 859: This should be fatal, use FAIL

tests/unittests/media/client/main/mediaPipeline/SetVolumeTest.cpp
```diff
@@ -58,9 +58,11 @@ TEST_F(RialtoClientMediaPipelineSetVolumeTest, setVolumeWithNoEaseTypeAndDuratio
 {
     EXPECT_CALL(*m_mediaPipelineIpcMock, setVolume(m_kTargetVolume, m_kVolumeDuration, m_kEaseType)).WillOnce(Return(true));
 
-    EXPECT_EQ(m_mediaPipeline->setVolume(m_kTargetVolume), true);
-}
+    auto mediaPipeline = dynamic_cast<IMediaPipeline *>(m_mediaPipeline.get());
```
skyfroxl, line 61: Use dynamic_pointer_cast and keep it as a shared_ptr

skyfroxl (dismissed):

media/server/gstplayer/source/tasks/generic/SetVolume.cpp
```diff
@@ -35,15 +43,75 @@ SetVolume::~SetVolume()
...
     {
         RIALTO_SERVER_LOG_ERROR("Setting volume failed. Pipeline is NULL");
         return;
     }
-    m_gstWrapper->gstStreamVolumeSetVolume(GST_STREAM_VOLUME(m_context.pipeline), GST_STREAM_VOLUME_FORMAT_LINEAR,
-                                           m_volume);
+    bool isNormalVolumeChange = (m_volumeDuration == 0);
+    GstElement *audioSink = m_player.getSink(firebolt::rialto::MediaSourceType::AUDIO);
+
+    if (!isNormalVolumeChange && audioSink &&
```
aczs, line 74: I would change the variable name to isImmediateVolumeChange
I would process the immediate volume change as first one. In my opinion it will be the one, which will be the most frequently used and the code will be more readable.

if (isImmediateVolumeChange)
...
else if (audioSink &&  m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(audioSink), "audio-fade"))
...
else if (m_rdkGstreamerUtilsWrapper->isSocAudioFadeSupported())
..
else
WRN log

aczs (approved):

skyfroxl (approved):

PR #330, opened 2024-09-13
Cache the immediate-output property until the pipeline is created

skyproudmanp:
Summary: Cache requests to set the immediate-output property until the pipeline is created. This needs to happen on both RialtoServer and rialto-gstreamer clients
Type: Fix  
Test Plan: UT, CT, native_build  
Jira: RIALTO-597


media/server/gstplayer/include/GstGenericPlayer.h
```diff
@@ -304,6 +307,24 @@ class GstGenericPlayer : public IGstGenericPlayer, public IGstGenericPlayerPriva
      */
     std::unique_ptr<IGstProtectionMetadataHelper> m_protectionMetadataWrapper;
 };
+
```
skywojciechowskim, line 310: new class -> new file

media/server/gstplayer/include/GenericPlayerContext.h
```diff
@@ -34,6 +34,12 @@
 namespace firebolt::rialto::server
 {
 constexpr double kNoPendingPlaybackRate{0.0};
+enum class PendingBool
```
skywojciechowskim, line 37: std::optional?

media/server/gstplayer/source/tasks/generic/SetImmediateOutput.cpp.rej
```diff
@@ -0,0 +1,51 @@
+--- media/server/gstplayer/source/tasks/generic/SetImmediateOutput.cpp
```
aczs, line 1: You dont want to add this file

media/server/gstplayer/include/GstGenericPlayer.h
```diff
@@ -167,6 +168,8 @@ class GstGenericPlayer : public IGstGenericPlayer, public IGstGenericPlayerPriva
     GstElement *getDecoder(const MediaSourceType &mediaSourceType) override;
 
 private:
+    friend class ActualSink;
```
skywojciechowskim, line 171: if I see correctly getSinkChildIfAutoVideoSink/getSinkChildIfAutoAudioSink will be used in a new class only, so please remove that friend and move those methods to ActualSink class

skywojciechowskim: In general, we should avoid using "friend" keyword, because usually it means, that it's something wrong with class interface.

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -478,33 +478,26 @@ bool GstGenericPlayer::setImmediateOutput(const MediaSourceType &mediaSourceType
...
     return true;
 }
 
 bool GstGenericPlayer::getImmediateOutput(const MediaSourceType &mediaSourceType, bool &immediateOutputRef)
 {
     bool returnValue{false};
-    GstElement *sink = getSink(mediaSourceType);
-    if (sink)
+    ActualSink sink(*this, mediaSourceType, m_gstWrapper);
+    if (sink.get())
```
skywojciechowskim, line 490: if we have a ActualSink class, maybe it's a good idea to add a method there -> setImmediateOutput. It would look better imo

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -517,21 +510,14 @@ bool GstGenericPlayer::getImmediateOutput(const MediaSourceType &mediaSourceType
 bool GstGenericPlayer::getStats(const MediaSourceType &mediaSourceType, uint64_t &renderedFrames, uint64_t &droppedFrames)
 {
     bool returnValue{false};
-    GstElement *sink = getSink(mediaSourceType);
-    if (sink)
+    ActualSink sink(*this, mediaSourceType, m_gstWrapper);
+    if (sink.get())
```
skywojciechowskim, line 514: same as for immediateOutput

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -950,32 +935,59 @@ void GstGenericPlayer::setEos(const firebolt::rialto::MediaSourceType &type)
 bool GstGenericPlayer::setVideoSinkRectangle()
 {
     bool result = false;
-    GstElement *videoSink = nullptr;
-    m_glibWrapper->gObjectGet(m_context.pipeline, "video-sink", &videoSink, nullptr);
-    if (videoSink)
+    ActualSink videoSink(*this, MediaSourceType::VIDEO, m_gstWrapper);
+    if (videoSink.get())
```
skywojciechowskim, line 939: as above :)

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -950,32 +935,59 @@ void GstGenericPlayer::setEos(const firebolt::rialto::MediaSourceType &type)
...
     return result;
 }
 
+bool GstGenericPlayer::setImmediateOutput()
+{
+    bool result{false};
+    if (m_context.pendingImmediateOutputForVideo != PendingBool::NOT_PENDING)
+    {
+        ActualSink sink(*this, MediaSourceType::VIDEO, m_gstWrapper);
+        if (sink.get())
```
skywojciechowskim, line 965: as above ;)

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -1205,20 +1217,18 @@ bool GstGenericPlayer::setSync(bool sync)
 bool GstGenericPlayer::getSync(bool &sync)
 {
     bool returnValue{false};
-    GstElement *sink = getSink(MediaSourceType::AUDIO);
-    if (sink)
+    ActualSink sink(*this, MediaSourceType::AUDIO, m_gstWrapper);
+    if (sink.get())
```
skywojciechowskim, line 1221: as above ;)

media/server/gstplayer/include/GenericPlayerContext.h
```diff
@@ -130,6 +130,12 @@ struct GenericPlayerContext
      */
     double pendingPlaybackRate{kNoPendingPlaybackRate};
 
+    /**
+     * @brief Pending immediate output for MediaSourceType::VIDEO, it's possible we'll have to add
```
skyfroxl, line 134: I would remove the bit 'it's possible we'll have to add a "pendingImmediateOutputForAudio" in the future.'

I don't think immediate-output is supported on any of the Audio sinks, if it ever is we can sort it then.

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -387,10 +387,11 @@ bool GstGenericPlayer::getPosition(std::int64_t &position)
     return true;
 }
 
-GstElement *GstGenericPlayer::getSink(const MediaSourceType &mediaSourceType)
+GstElement *GstGenericPlayer::getSink(GstObject *&pointerToUnref, const MediaSourceType &mediaSourceType) const
```
skyfroxl, line 390: How about instead of adding this complexity, we make it so the returned sink always has to be unrefed.

We could do this by increasing the reference of autosink (if there is one) using gst_object_ref in getSinkChildIfAutoVideoSink.

skyfroxl (approved):

skywojciechowskim (approved):

PR #334, opened 2024-09-26
BufferingLimit and UseBuffering and (Video)SyncStreamMode properties support added

skywojciechowskim:
Summary: BufferingLimit and UseBuffering and VideoSyncStreamMode properties support added
Type: Feature
Test Plan: UT/ CT
Jira: RIALTO-639


skyshaha1 (approved):

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -498,6 +498,62 @@ GstElement *GstGenericPlayer::getDecoder(const MediaSourceType &mediaSourceType)
     return nullptr;
 }
 
+GstElement *GstGenericPlayer::getFilter(const MediaSourceType &mediaSourceType)
```
aczs, line 501: This functions gets parser, not filter (from gstreamer type classification perspective). 
Have you considered using gstElementFactoryListGetElements? That will give you a list of of factories that match the given type. Then, you would be able to iterate through them to find the property.

aczs, line 501: This functions searches for parser, not filter from gstreamer's classification perspective.
What if you would have multiple PARSER | MEDIA_SOURCE elements in the pipeline? It would find only the first element that matches the type. Maybe it would be better to find element that matches the given type AND has to searches property? OR function, which returns all elements of given type in the pipeline.

skywojciechowskim: there is no such thing like TYPE_FILTER in gstreamer. Filters have Codec/Parser type.

skywojciechowskim: function name changed to getParser

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -1523,14 +1653,73 @@ void GstGenericPlayer::processAudioGap(int64_t position, uint32_t duration, int6
...
+{
+    if (m_workerThread)
+    {
+        m_workerThread->enqueueTask(m_taskFactory->createSetUseBuffering(m_context, *this, useBuffering));
+    }
+}
+
+bool GstGenericPlayer::getUseBuffering(bool &useBuffering)
+{
+    if (m_context.playbackGroup.m_curAudioDecodeBin)
```
skyfroxl, line 1701: Using m_context.playbackGroup.m_curAudioDecodeBin on the main thread may not be thread safe

skywojciechowskim: m_curAudioDecodeBin changed to atomic. Additionally, property mutex introduced to protect properties in get/set methods

skyfroxl (approved):

skyshaha1 (approved):

PR #336, opened 2024-10-10
Add a volume property to the web audio

skyproudmanp:
Summary: Implement a "volume" property on Web Audio to set/attenuate the volume before mixing with the Media Player & TTS audio output (the volume of the latter will not be affected)
Type: Feature
Test Plan: UT, CT and test on XiOne US 
Jira: RIALTO-595

### DIFF
--- a/media/server/gstplayer/source/GstWebAudioPlayer.cpp
+++ b/media/server/gstplayer/source/GstWebAudioPlayer.cpp
@@ -239,37 +239,43 @@ GstElement *GstWebAudioPlayer::createAutoSink()
 
 bool GstWebAudioPlayer::linkElementsToSrc(GstElement *sink)
 {
-    GstElement *convert = NULL;
-    GstElement *resample = NULL;
-
-    convert = m_gstWrapper->gstElementFactoryMake("audioconvert", NULL);
+    GstElement *convert = m_gstWrapper->gstElementFactoryMake("audioconvert", NULL);
     if (!convert)
     {
         RIALTO_SERVER_LOG_ERROR("Failed create the audioconvert");
         return false;
     }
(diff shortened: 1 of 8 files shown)


media/server/gstplayer/source/GstWebAudioPlayer.cpp
```diff
@@ -239,38 +239,44 @@ GstElement *GstWebAudioPlayer::createAutoSink()
...
     if (!resample)
     {
         RIALTO_SERVER_LOG_ERROR("Failed create the audioresample");
         return false;
     }
+    GstElement *volume = m_gstWrapper->gstElementFactoryMake("volume", NULL);
+    if (!volume)
+    {
+        RIALTO_SERVER_LOG_ERROR("Failed create the volume");
+        return false;
```
skyfroxl, line 258: No coverage for error case, can we add a test

media/server/gstplayer/source/GstWebAudioPlayer.cpp
```diff
@@ -239,38 +239,44 @@ GstElement *GstWebAudioPlayer::createAutoSink()
...
     if (!convert)
     {
         RIALTO_SERVER_LOG_ERROR("Failed create the audioconvert");
         return false;
     }
-    resample = m_gstWrapper->gstElementFactoryMake("audioresample", NULL);
+    GstElement *resample = m_gstWrapper->gstElementFactoryMake("audioresample", NULL);
     if (!resample)
     {
         RIALTO_SERVER_LOG_ERROR("Failed create the audioresample");
```
skywojciechowskim, line 251: (i'm not sure!) shouldn't we unref convert here?

skyfroxl (dismissed):

skyshaha1 (approved):

skyfroxl (approved):

PR #344, opened 2024-11-08
Position reporting and getVolume performance improvements.

skywojciechowskim:
Summary: Position reporting and getVolume performance improvements.
Type: Fix
Test Plan: UT/ CT
Jira: RIALTO-667

### DIFF
--- a/media/server/gstplayer/source/GstGenericPlayer.cpp
+++ b/media/server/gstplayer/source/GstGenericPlayer.cpp
@@ -1467,7 +1467,7 @@ bool GstGenericPlayer::getVolume(double &currentVolume)
     }
 
     GstElement *sink{getSink(MediaSourceType::AUDIO)};
-    if (m_context.fadeVolumeSet && sink &&
+    if (m_context.audioFadeEnabled && sink &&
         m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(sink), "fade-volume"))
     {
         gint fadeVolume{-100};
--- a/media/server/gstplayer/source/GstGenericPlayer.cpp
+++ b/media/server/gstplayer/source/GstGenericPlayer.cpp
@@ -1467,7 +1467,8 @@ bool GstGenericPlayer::getVolume(double &currentVolume)
     }
 
     GstElement *sink{getSink(MediaSourceType::AUDIO)};
-    if (sink && m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(sink), "fade-volume"))
+    if (m_context.fadeVolumeSet && sink &&
+        m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(sink), "fade-volume"))
     {
         gint fadeVolume{-100};
         m_glibWrapper->gObjectGet(sink, "fade-volume", &fadeVolume, NULL);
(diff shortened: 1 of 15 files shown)


media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -1467,7 +1467,8 @@ bool GstGenericPlayer::getVolume(double &currentVolume)
     }
 
     GstElement *sink{getSink(MediaSourceType::AUDIO)};
-    if (sink && m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(sink), "fade-volume"))
+    if (m_context.audioFadeEnabled && sink &&
+        m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(sink), "fade-volume"))
```
skyproudmanp, line 1471: If speed is an issue then wouldn't it be quicker to check this once...
 m_glibWrapper->gObjectClassFindProperty(G_OBJECT_GET_CLASS(sink), "fade-volume")
 ...before setting m_context.audioFadeEnabled (this is set in two places, and the property is already checked for in one of these two places. We'd just need to add a check before the other location too, and perhaps add a comment to the header file to say that the existence of the property must be verified before setting the variable)

skywojciechowskim: I don't agree, in setVolume, where audioFadeEnabled is set, "audio-fade property" is checked. Here we check "fade-volume", so it's different property. gObjectClassFindProperty is fast, on broadcom it always returns true. Slow operation, that we want to avoid when audio-fade is disabled is: gObjectSet(audioSink, "audio-fade", fadeStr, nullptr);

skywojciechowskim: Unfortunately it's not possible to check if fade-volume is present in else if in set volume - sink may be null there, so code has to stay as it is now

skyproudmanp (approved):

media/server/main/source/MainThread.cpp
```diff
@@ -160,12 +160,31 @@ void MainThread::enqueueTaskAndWait(uint32_t clientId, Task task)
...
+    newTask->clientId = clientId;
+    newTask->task = task;
+    newTask->mutex = std::make_unique<std::mutex>();
+    newTask->cv = std::make_unique<std::condition_variable>();
+
+    {
+        std::unique_lock<std::mutex> lockTask(*(newTask->mutex));
+        {
+            std::unique_lock<std::mutex> lockQueue(m_taskQueueMutex);
+            m_taskQueue.push_front(newTask);
```
aczs, line 183: Shouldn't we use use priority queue instead? Right now it's only one task with higher priority, so it's not a problem, but if we would like to prioritise more tasks in the future, then prio queue would be more suitable in my opinion.

aczs (approved):

media/server/main/source/MainThread.cpp
skyproudmanp: Since we only have 2 levels of priority then I think that priority queue isn't necessary (for ANY number of tasks). A double ended queue is simple and as fast as it could be?

(Priority queue just optimises fetching the largest or smallest from a collection, so we'd have to make every task aware of it's own priority for sorting purposes)

aczs: The problem isn't that tasks with different priorities would be mixed up, but that high priority tasks would be scheduled in LIFO and low prio tasks would be scheduled in FIFO. Using LIFO for high priority tasks could lead to nasty race conditions.
I already talked with Marcin about that and we are guarded by blocked threads, but anyway I would change to prio queue for readability.

skywojciechowskim: as discussed on slack, it's not possible

PR #345, opened 2024-11-12
Require Audio Easing support on RialtoMSEAudioSink for Netflix (comments added around the fade-volume prop ) 

skyshaha1:
Summary: Require Audio Easing support on RialtoMSEAudioSink for Netflix - comments added around fade volume prop
Type: Fix
Test Plan: UT/ CT
Jira: RIALTO-593

### DIFF
--- a/media/server/gstplayer/source/GstGenericPlayer.cpp
+++ b/media/server/gstplayer/source/GstGenericPlayer.cpp
@@ -1466,27 +1466,33 @@ bool GstGenericPlayer::getVolume(double &currentVolume)
         return false;
     }
 
+    // NOTE: No gstreamer documentation for "fade-volume" could be found at the time this code was written.
+    // Therefore the author performed several tests on a supported platform (Flex2) to determine the behaviour of this property.
+    // The code has been written to be backwardly compatible on platforms that don't have this property.
+    // The observed behaviour was:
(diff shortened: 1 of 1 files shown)


aczs (approved):

media/server/gstplayer/source/GstGenericPlayer.cpp
```diff
@@ -1467,23 +1467,26 @@ bool GstGenericPlayer::getVolume(double &currentVolume)
     }
```
skyproudmanp, line 1468: The comments below just say what the code does (which doesn't need extra explanation IMO)

I'd recommend adding a single comment above all of this code that explains something like this (please feel free to correct/ adapt this):-

// NOTE: No gstreamer documentation for "fade-volume" could be found at the time this code was written.
// Therefore the author performed several tests on a supported platform (flex2) to determine the behaviour
// of this property. Also the code has been written to be backwardly compatible on platforms that
// don't have this property. The observed behaviour was:-
//  - if the returned fade volume is negative then fade-volume is not active. In this case the usual technique to find volume works
//  - if the returned fade volume is positive then fade-volume is active. In this case the returned fade volume is like the "master volume" and directly returns the current volume level 0=min to 100=max (and the pipeline's current volume level is meaningless and doesn't contribute in this case)

skyproudmanp (approved):

aczs (approved):

PR #348, opened 2024-11-21
Config file: camelCase naming, support for soc config file added

skywojciechowskim:
Summary: Added camel case naming in rialto config file. Added support for soc config file.
Type: Feature
Test Plan: UT/ Fullstack
Jira: RIALTO-702 RIALTO-709 RIALTO-710

### DIFF
--- a/serverManager/service/CMakeLists.txt
+++ b/serverManager/service/CMakeLists.txt
@@ -68,6 +68,7 @@ if( RIALTO_ENABLE_CONFIG_FILE )
     add_compile_definitions(RIALTO_ENABLE_CONFIG_FILE)
     add_compile_definitions(RIALTO_CONFIG_PATH="${RIALTO_CONFIG_FILE_DIR}/rialto-config.json")
     add_compile_definitions(RIALTO_CONFIG_OVERRIDES_PATH="${RIALTO_CONFIG_OVERRIDES_FILE_DIR}/rialto-config-overrides.json")
+    add_compile_definitions(RIALTO_CONFIG_SOC_PATH="${RIALTO_CONFIG_FILE_DIR}/rialto_soc.json")
 
     target_link_libraries(
         RialtoServerManager
--- a/serverManager/service/include/ConfigHelper.h
+++ b/serverManager/service/include/ConfigHelper.h
@@ -50,11 +50,14 @@ public:
 #ifdef RIALTO_ENABLE_CONFIG_FILE
 private:
     void readConfigFile(const std::string &filePath);
+    void mergeEnvVariables();
 #endif // RIALTO_ENABLE_CONFIG_FILE
 
 private:
     std::unique_ptr<IConfigReaderFactory> m_configReaderFactory;
     std::map<std::string, std::string> m_sessionServerEnvVars;
+    std::map<std::string, std::string> m_jsonEnvVars;
+    std::map<std::string, std::string> m_jsonExtraEnvVars;
     std::string m_sessionServerPath;
     std::chrono::milliseconds m_sessionServerStartupTimeout;
     std::chrono::seconds m_healthcheckInterval;
--- a/serverManager/service/include/ConfigHelper.h
+++ b/serverManager/service/include/ConfigHelper.h
@@ -56,8 +56,8 @@ private:
 private:
     std::unique_ptr<IConfigReaderFactory> m_configReaderFactory;
     std::map<std::string, std::string> m_sessionServerEnvVars;
-    std::map<std::string, std::string> m_jsonEnvVars;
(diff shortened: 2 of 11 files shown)


serverManager/service/source/ConfigReader.cpp
```diff
@@ -75,33 +88,71 @@ void ConfigReader::parseEnvironmentVariables(std::shared_ptr<firebolt::rialto::w
...
+                m_extraEnvVars.emplace_back(envVarsJson->at(index)->asString());
+            }
+        }
+    }
+}
+
 void ConfigReader::parseSessionServerPath(std::shared_ptr<firebolt::rialto::wrappers::IJsonValueWrapper> root)
 {
-    if (root->isMember("session_server_path") && root->at("session_server_path")->isString())
+    if (root->isMember("sessionServerPath") && root->at("sessionServerPath")->isString())
```
skyshaha1, line 109: This logic is repeated across multiple parsing methods, but with different key names, value types and variables, so maybe we could create a template helper to reduce the duplicated code making it cleaner - the reason I thought of a template is for type safety - but this might be unnecessary & an over optimisation in this case, lmk what you think

skywojciechowskim: done

serverManager/service/CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ if( RIALTO_ENABLE_CONFIG_FILE )
     add_compile_definitions(RIALTO_ENABLE_CONFIG_FILE)
     add_compile_definitions(RIALTO_CONFIG_PATH="${RIALTO_CONFIG_FILE_DIR}/rialto-config.json")
     add_compile_definitions(RIALTO_CONFIG_OVERRIDES_PATH="${RIALTO_CONFIG_OVERRIDES_FILE_DIR}/rialto-config-overrides.json")
+    add_compile_definitions(RIALTO_CONFIG_SOC_PATH="${RIALTO_CONFIG_FILE_DIR}/rialto_soc.json")
```
skyproudmanp, line 71: Should this be ".../rialto-soc.json" (dash not underscore) to be consistent with the file naming conventions on lines 69 & 70

skyshaha1 (approved):

serverManager/service/include/ConfigHelper.h
```diff
@@ -50,11 +50,14 @@ class ConfigHelper
 #ifdef RIALTO_ENABLE_CONFIG_FILE
 private:
     void readConfigFile(const std::string &filePath);
+    void mergeEnvVariables();
 #endif // RIALTO_ENABLE_CONFIG_FILE
 
 private:
     std::unique_ptr<IConfigReaderFactory> m_configReaderFactory;
     std::map<std::string, std::string> m_sessionServerEnvVars;
+    std::map<std::string, std::string> m_jsonEnvVars;
```
skyproudmanp, line 59: Just a suggestion....
I'm wondering if there's a better name, because I presume that these settings are no longer in json format when stored in the map?
Perhaps envVarsFromConfigFile ? (but feel free to decide)

serverManager/service/source/ConfigHelper.cpp
```diff
@@ -126,14 +129,18 @@ void ConfigHelper::readConfigFile(const std::string &filePath)
         return;
     }
 
-    std::map<std::string, std::string> envVariables{convertToMap(configReader->getEnvironmentVariables())};
-    for (const auto &[name, value] : envVariables)
+    // Always override env variables when present in "more important" file
+    // (envVariables + extraEnvVariables from less important file will be wiped if envVariables are present)
```
skyproudmanp, line 133: I like this comment.
Therefore, would "lowerPriorityEnvVariables" be a better name than "extraEnvVariables" ?

serverManager/service/CMakeLists.txt
skywojciechowskim: it can't be. Comcast decided to have rialto_soc.json

serverManager/service/include/ConfigHelper.h
skywojciechowskim: If you know better name, suggest one here :)

serverManager/service/source/ConfigHelper.cpp
skywojciechowskim: again, comcast decided to have extraEnvVariables, we can't do anything with it

serverManager/service/include/ConfigHelper.h
skywojciechowskim: changed.

skyproudmanp (approved):

skyshaha1 (approved):

PR #349, opened 2024-11-22
Audio source switch dedicated method

skywojciechowskim:
Summary: Audio source switch dedicated method added
Type: Fix
Test Plan: UT/ CT, Fullstack
Jira: RIALTO-635


media/server/gstplayer/source/Utils.cpp
```diff
@@ -133,4 +136,65 @@ std::string getUnderflowSignalName(const firebolt::rialto::wrappers::IGlibWrappe
     }
     return "";
 }
+
+GstCaps *createCapsFromMediaSource(const std::shared_ptr<firebolt::rialto::wrappers::IGstWrapper> &gstWrapper,
```
skyshaha1, line 140: There is a lot of repeated logic here, maybe we could refactor the if-else logic into a function  to encapsulate the decision making for the createCapsFromMediaSource function. We could maybe introduce a function to the config types and then update the createCapsFromMediaSource to use that function - just to reduce duplication of code and making it easier to read

skywojciechowskim: it's an implementation moved from AttachSource. We can create a new task to refactor it, if you think, that it's needed :)

skyshaha1: Ahh yeh I see, I still think it would simplify the logic in createCapsFromMediaSource making it easier to read, and it would make it easier to add new media types (in the future)

skyshaha1: might also need to adjust the ut as well then

tests/componenttests/server/tests/mediaPipeline/AudioSourceSwitchTest.cpp
```diff
@@ -195,4 +229,111 @@ TEST_F(AudioSourceSwitchTest, SwitchAudioSource)
...
 }
+
+/*
+ * Component Test: Switch audio source procedure test
+ * Test Objective:
+ *  Test the audio switch procedure
+ *
+ * Sequence Diagrams:
+ *  Rialto Dynamic Audio Stream Switching
+ *   - https://wiki.rdkcentral.com/pages/viewpage.action?spaceKey=ASP&title=Rialto+Dynamic+Audio+Stream+Switching
```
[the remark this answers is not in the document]
skywojciechowskim: Yes, as I wrote on slack, I need to update docs too. I'll do it now.

tests/componenttests/server/tests/mediaPipeline/AudioSourceSwitchTest.cpp
```diff
@@ -195,4 +229,111 @@ TEST_F(AudioSourceSwitchTest, SwitchAudioSource)
...
+    willSetupAndAddSource(&m_audioAppSrc);
+    willSetupAndAddSource(&m_videoAppSrc);
+    willFinishSetupAndAddSource();
+    indicateAllSourcesAttached();
+
+    // Step 4: Switch Audio Source
+    willSwitchAudioSource();
+    switchAudioSource();
+
+    // Step 5: Remove sources
```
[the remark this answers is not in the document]
skywojciechowskim: because I forgot to uncomment it :p thanks for spotting :)

skyfroxl (approved):

PR #354, opened 2025-01-10
Subtitle set position redesigned. Fix subtitle position reporting

skywojciechowskim:
Summary: Subtitle set position redesigned. Fix subtitle position reporting
Type: Fix
Test Plan: UT/ CT, Fullstack
Jira: RIALTO-705

### DIFF
--- a/media/server/gstplayer/source/GstTextTrackSink.cpp
+++ b/media/server/gstplayer/source/GstTextTrackSink.cpp
@@ -399,7 +399,7 @@ static void gst_rialto_text_track_sink_set_property(GObject *object, guint propI
     }
 }
 
-static gboolean gst_rialto_text_track_sink_query(GstElement *element, GstQuery *query)
+static gboolean gst_rialto_text_track_sink_query(GstElement *element, GstQuery *query) // NOLINT(build/function_format)
 {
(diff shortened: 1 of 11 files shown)


media/server/gstplayer/source/GstTextTrackSink.cpp
```diff
@@ -368,12 +379,57 @@ static void gst_rialto_text_track_sink_set_property(GObject *object, guint propI
...
+    {
+    case GST_QUERY_POSITION:
+    {
+        GstFormat fmt;
+        gst_query_parse_position(query, &fmt, NULL);
+        switch (fmt)
+        {
+        case GST_FORMAT_TIME:
+        {
+            // Thunder ITextTrack does not provide getPosition API so we are unable to determine current position
```
aczs, line 416: In case of position query, we shouldn't return TextTrack's position even if it would supply such option.
That would confuse gstreamer's pipeline.

aczs (dismissed):

media/server/gstplayer/source/GstTextTrackSink.cpp
skywojciechowskim: comment modified

aczs (approved):

PR #363, opened 2025-02-13
NeedMediaData timers changed for low latency

skywojciechowskim:
Summary: NeedMediaData timers changed for low latency
Type: Fix
Test Plan: UT
Jira: ENTDAI-459

### DIFF
--- a/media/server/main/source/MediaPipelineServerInternal.cpp
+++ b/media/server/main/source/MediaPipelineServerInternal.cpp
@@ -32,7 +32,6 @@
 
 namespace
 {
-constexpr std::chrono::milliseconds kNeedMediaDataResendTimeMs{100};
 const char *toString(const firebolt::rialto::MediaSourceStatus &status)
 {
     switch (status)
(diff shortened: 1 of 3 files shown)


media/server/main/source/MediaPipelineServerInternal.cpp
```diff
@@ -1006,6 +1007,8 @@ bool MediaPipelineServerInternal::setLowLatencyInternal(bool lowLatency)
         RIALTO_SERVER_LOG_ERROR("Failed to set low latency - Gstreamer player has not been loaded");
         return false;
     }
+    m_IsLowLatencyAudioPlayer = lowLatency;
```
skyproudmanp, line 1010: Just a general comment...
  setLowLatency and setLowLatencyInternal
should really be called...
  setLowLatencyForAudio and setLowLatencyForAudioInternal

I can see that this function is only called for audio streams (see line of code below)
  media/server/gstplayer/source/tasks/generic/SetupElement.cpp:272
  
Given the limited rialto resource I'll approve this anyway. Perhaps improved names could be done in the future.

--

Also setImmediateOutput should be called setImmediateOutputForVideo. And, I am responsible for this non-descriptive name :-)

aczs (approved):

PR #388, opened 2025-06-18
Valgrind on native

yurii-yakubin:
Summary: Valgrind on native
Type: Feature
Test Plan: UT/CT, Fullstack
Jira: ENTDAI-464

### DIFF
--- a/media/server/gstplayer/source/GstInitialiser.cpp
+++ b/media/server/gstplayer/source/GstInitialiser.cpp
@@ -30,11 +30,13 @@ GstInitialiser::~GstInitialiser()
         m_thread.join();
     }
 
+#ifdef FREE_MEM_BEFORE_EXIT
     if (m_gstWrapper)
     {
         m_gstWrapper->gstDeinit();
         m_gstWrapper.reset();
     }
+#endif
 }
 
 IGstInitialiser &IGstInitialiser::instance()


wrappers/interface/IGstWrapper.h
```diff
@@ -78,6 +78,12 @@ class IGstWrapper
      */
     virtual void gstInit(int *argc, char ***argv) = 0;
 
+    /**
+     * @brief Deinitalise gstreamer.
+     *
+     */
+    virtual void gstDeinit() = 0;
```
skywojciechowskim, line 85: dead code - you don't call this function

media/server/service/source/main.cpp
```diff
@@ -47,13 +49,25 @@ int main(int argc, char *argv[])
 
     firebolt::rialto::server::IGstInitialiser::instance().initialise(&argc, &argv);
 
-    auto appSessionServer =
-        firebolt::rialto::server::IApplicationSessionServerFactory::getFactory()->createApplicationSessionServer();
-
-    if (!appSessionServer->init(argc, argv))
     {
-        return EXIT_FAILURE;
+        // Creation of this variable in a local scope ensures Rialto
```
skywojciechowskim, line 53: you probably wanted to call deinit at the end of main, but you don't call it. So is that change still needed?

wrappers/interface/IGstWrapper.h
skywojciechowskim: Well, diff shows, that you added it :p Is gst_deinit additional call needed to fix any of valgrind issues?

media/server/service/source/main.cpp
yurii-yakubin: Added gstDeinit call into GstInitialiser destructor

aczs: Do we really need the inner scope here? Isn't it guaranteed that destructor of a local variable in main will be called before destructor of a singleton?

aczs (approved):

media/server/service/source/main.cpp
yurii-yakubin: Yes, a local scope is also not needed.

skywojciechowskim (dismissed):

media/server/service/source/main.cpp
```diff
@@ -55,5 +57,11 @@ int main(int argc, char *argv[])
         return EXIT_FAILURE;
     }
     appSessionServer->startService();
+
+#ifdef FREE_MEM_BEFORE_EXIT
```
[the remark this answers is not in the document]
yurii-yakubin: done

aczs (approved):

PR #463, opened 2026-03-16
Fix memory leaks in Rialto master

Koky2701:
Fix memory leaks in Rialto master

Summary: Finding potential memory leak points and adding fixes
Type: Fix
Test Plan: UT/CT, Fullstack
Jira: RDKEMW-15495


media/server/ipc/include/ControlModuleService.h
```diff
@@ -62,7 +62,9 @@ class ControlModuleService : public IControlModuleService
 private:
     service::IPlaybackService &m_playbackService;
     service::IControlService &m_controlService;
-    std::map<std::shared_ptr<::firebolt::rialto::ipc::IClient>, std::set<int>> m_controlIds;
+    std::map<std::weak_ptr<::firebolt::rialto::ipc::IClient>, std::set<int>,
```
skywojciechowskim: In my opinion this change will break the functionality.
First of all, Copilot comment is a false positive - it claims that we will have cycles, because there is an "exportService(shared_from_this())" called. In the reality, new client is added in void SessionManagementServer::onClientConnected(const std::shared_ptr<::firebolt::rialto::ipc::IClient> &client) method. If you look when this method is called, you will notice, that it's a callback invoked in IpcServerImpl.cpp void ServerImpl::processNewConnection(uint64_t socketId). As you can see, new shared ptr is created there and it is passed later to the ControlModuleService. So if you use the weak_ptr instead of shared_ptr, reference count of this shared_ptr may drop to 0.

skywojciechowskim (dismissed):

skywojciechowskim (approved):

DouglasAdler (approved):
Looks good, please examine the CoPilot comments and resolve them before merging

PR #487, opened 2026-04-21
Fixed initialisation of playbackGroup elements in GenericPlayerContext. Added linking of typefind and the parser after the audio switch

smudri85:
Summary: Fixed initialisation of playbackGroup elements in GenericPlayerContext. Added linking of typefind and the parser after the audio switch
Type: Feature
Test Plan: UT/CT, Fullstack
Jira: RDKEMW-16807


Koky2701 (approved):
Look's good!

PR #578, opened 2026-08-05
RDKEMW-22812: Add persistent DRM session support in Rialto

varatharajan568:
RDKEMW-22812: Add persistent DRM session support in Rialto

Reason for change: Implement persistent-license session support in Rialto-OCDM for Widevine and PlayReady DRM. This enables applications to create, store, restore, and reuse persistent DRM sessions across application restarts in compliance with EME persistent-license requirements.
Test Procedure: Verify persistent-license session creation for Widevine & PlayReady DRM

### DIFF
--- a/media/server/main/include/IMediaKeySession.h
+++ b/media/server/main/include/IMediaKeySession.h
@@ -83,15 +83,15 @@ public:
      *
      * @param[in]  initDataType : The init data type.
      * @param[in]  initData     : The init data.
-    * @param[in]  cdmData      : Optional CDM data.
+     * @param[in]  cdmData      : Optional CDM data.
      * @param[in]  ldlState     : The Limited Duration License state. Most of key systems do not need this parameter.
      *
      * @retval an error status.
      */
-    virtual MediaKeyErrorStatus generateRequest(InitDataType initDataType, const std::vector<uint8_t> &initData,
-                                                const std::vector<uint8_t> &cdmData = std::vector<uint8_t>{},
-                                                const LimitedDurationLicense &ldlState =
-                                                    LimitedDurationLicense::NOT_SPECIFIED) = 0;
+    virtual MediaKeyErrorStatus
+    generateRequest(InitDataType initDataType, const std::vector<uint8_t> &initData,
+                    const std::vector<uint8_t> &cdmData = std::vector<uint8_t>{},
+                    const LimitedDurationLicense &ldlState = LimitedDurationLicense::NOT_SPECIFIED) = 0;
     /**
      * @brief Loads the existing key session.
      *
--- a/media/server/main/include/IMediaKeySession.h
+++ b/media/server/main/include/IMediaKeySession.h
@@ -83,12 +83,15 @@ public:
      *
      * @param[in]  initDataType : The init data type.
      * @param[in]  initData     : The init data.
+    * @param[in]  cdmData      : Optional CDM data.
      * @param[in]  ldlState     : The Limited Duration License state. Most of key systems do not need this parameter.
      *
      * @retval an error status.
      */
     virtual MediaKeyErrorStatus generateRequest(InitDataType initDataType, const std::vector<uint8_t> &initData,
-                                                const LimitedDurationLicense &ldlState) = 0;
+                                                const std::vector<uint8_t> &cdmData = std::vector<uint8_t>{},
+                                                const LimitedDurationLicense &ldlState =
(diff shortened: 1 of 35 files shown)


media/server/main/include/IMediaKeySession.h
```diff
@@ -83,12 +83,15 @@ class IMediaKeySession
...
      * @param[in]  initDataType : The init data type.
      * @param[in]  initData     : The init data.
+    * @param[in]  cdmData      : Optional CDM data.
      * @param[in]  ldlState     : The Limited Duration License state. Most of key systems do not need this parameter.
      *
      * @retval an error status.
      */
     virtual MediaKeyErrorStatus generateRequest(InitDataType initDataType, const std::vector<uint8_t> &initData,
-                                                const LimitedDurationLicense &ldlState) = 0;
+                                                const std::vector<uint8_t> &cdmData = std::vector<uint8_t>{},
```
lukasz-winiarski-red, line 92: why do you need default values on server side?
I think default value is needed just for client - if he doesn't want to pass cdm data, but in client-server IPC communication that data would be always passed (even  if default), so default should not be needed on server side

media/server/main/include/MediaKeySession.h
```diff
@@ -72,7 +72,9 @@ class MediaKeySession : public IMediaKeySession, public firebolt::rialto::wrappe
     virtual ~MediaKeySession();
 
     MediaKeyErrorStatus generateRequest(InitDataType initDataType, const std::vector<uint8_t> &initData,
-                                        const LimitedDurationLicense &ldlState) override;
+                                        const std::vector<uint8_t> &cdmData = std::vector<uint8_t>{},
```
lukasz-winiarski-red, line 75: I believe you need default values just in interface, not in every inherited method

media/server/main/source/MediaKeySession.cpp
```diff
@@ -109,15 +109,15 @@ MediaKeySession::~MediaKeySession()
 }
 
 MediaKeyErrorStatus MediaKeySession::generateRequest(InitDataType initDataType, const std::vector<uint8_t> &initData,
+                                                     const std::vector<uint8_t> &cdmData,
                                                      const LimitedDurationLicense &ldlState)
 {
     RIALTO_SERVER_LOG_DEBUG("entry:");
-    if (LimitedDurationLicense::NOT_SPECIFIED != ldlState)
+     if (LimitedDurationLicense::NOT_SPECIFIED != ldlState)
```
lukasz-winiarski-red, line 116: alignment broken?

media/server/main/source/MediaKeySession.cpp
```diff
@@ -129,7 +129,12 @@ MediaKeyErrorStatus MediaKeySession::generateRequest(InitDataType initDataType,
     {
         initOcdmErrorChecking();
 
-        status = m_ocdmSession->constructSession(m_kSessionType, initDataType, &initData[0], initData.size());
+        const uint8_t *initDataPtr = initData.empty() ? nullptr : initData.data();
+        const uint8_t *cdmDataPtr = cdmData.empty() ? nullptr : cdmData.data();
+        const uint32_t cdmDataSize = cdmData.size();
```
lukasz-winiarski-red, line 134: why do you need all those local vars? calling init.data.data() is safe for empty vector as well

media/server/main/source/MediaKeySession.cpp
```diff
@@ -129,7 +129,12 @@ MediaKeyErrorStatus MediaKeySession::generateRequest(InitDataType initDataType,
...
+
+        status = m_ocdmSession->constructSession(m_kSessionType, initDataType, initDataPtr, initData.size(), cdmDataPtr,
```
lukasz-winiarski-red, line 136: I don't think that the logic here is correct:
for persistent license sessions you don't want any challenge generation (this is offline playback case) - so session id should be passed (via cdmData) - session should be constructed on OCDM side and that's it - then client is expected to call load

in current flow challenge generation will happen unconditionally in my understanding

media/server/main/source/MediaKeySession.cpp
lukasz-winiarski-red, line 81: Rialto design here seems problematic - creating session on OCDM side is delayed until generateRequest call, but generateRequest is not needed for some scenarios related to persistent licenses

lukasz-winiarski-red (changes requested):

PR #584, opened 2026-08-11
Fix sending additional GST segment after back to back flush calls

smudri85:
Summary: Fix sending additional GST segment after back to back flush calls.
Type: Fix
Test Plan: UT/CT, Fullstack
Jira: RDKEMW-23222


skywojciechowskim (approved):
Different version needs to be merged to later branches.
Accepted this one because of the time pressure.

